Issue #1, opened 2018-06-04
Build NLDAS hosting subset pipeline for domain


jread-usgs:
_could_ include intermediate data optimization and re-writes

jread-usgs:
Once we get past the auth part, we'll want `	dods://hydro1.gesdisc.eosdis.nasa.gov/dods/NLDAS_FORA0125_H.002
`

jread-usgs:
configure locally:
- [ ] .netrc
- [ ] .urs_cookies 
- [ ] .dodsrc

from: http://hdfeos.org/software/ncl_opendap_mls.php

jread-usgs:
If configured properly, this should run fine: 
```
nccopy -m 15m http://hydro1.sci.gsfc.nasa.gov/dods/NLDAS_FORA0125_H.002?lon[0:1:223],lat[0:1:163],time[0:1:15],pressfc[0:1:15][0:1:163][0:1:223] example.nc
```

jread-usgs:
[NLDAS data rods](https://hydro1.gesdisc.eosdis.nasa.gov/daac-bin/access/timeseries.cgi?variable=NLDAS:NLDAS_FORA0125_H.002:DSWRFsfc&location=NLDAS:X304-Y071&startDate=1979-01-02T00&endDate=1981-05-20T23&type=asc2) is another option, but it has a [water mask](https://ldas.gsfc.nasa.gov/nldas/asc/NLDASmask_UMDunified.asc) that will cause trouble for big lakes and potentially lakes close to the Great Lake shorelines.

jread-usgs:
Seems bypassing the GDP processing is a better solution, and avoiding a generic/reusable hosting of these timeseries data is practical at this point. Future needs for projects (e.g., 2WP) are expected to be covered in a more robust way w/ CONUS coverage.

Proposal:
1) timeseries driver files are written as `feather` files, which can be written to csv on demand when models are run. This format will be small locally, easy to open and append to, and readable/writeable by R and python
2) we'll ditch the intermediate netcdf file chunks. I don't see a good reason to keep them around, other than to avoid future downloads, and their size adds up
3) variables will be converted into input variables with proper units, but keep the native timestep (hourly). Downsampling to a less frequent timestep can be done on the fly as a pre-model step if necessary. Seeming increasingly likely we'll be running the model at an hourly timestep anyhow
[...]

PR #2, opened 2018-06-06
first part of NLDAS cubes workflow

jread-usgs:
small part of #1


0_config_rmk.yml
```diff
@@ -0,0 +1,31 @@
...
+    command: list(t0=I(0), t1 = I(12))
+    
+  nldas_time_stride:
+    command: c(I(10))
+    
+  nldas_cells:
+    command: data.frame(x=I(c(231,239,223)),y=I(c(135,135,141))) # will come from code or spatial data...
+    
+  nldas_cube:
+    command: nldas_cube_diff(nldas_cells, '0_config/in/nldas_cells.rds')
```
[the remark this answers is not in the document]
aappling-usgs: and would use `I()` around the 2nd arg, right? so that merely generating some feather files wouldn't force this command to come up with a new set of cubes?

0_config/code/nldas_cube_utils.R
```diff
@@ -0,0 +1,57 @@
+
+nldas_cube_diff <- function(new_cube_df, old_cube_df_filename){
+  old_cube_df <- readRDS(old_cube_df_filename)
+  diffed_cells <- dplyr::anti_join(new_cube_df, old_cube_df,  by = c('x','y')) 
+  
+  if (any(is.na(diffed_cells))){
+    stop('found NA(s) in NLDAS cell diff. Check ', old_cube_df_filename, ' and `new_cube_df` data')
+  }
+  cube_bounds <- diffed_cells %>% 
+    summarize(x0 = min(x), x1 = max(x), y0 = min(y), y1 = max(y))
```
aappling-usgs: looks like this is just doing a spatial diff, right? makes sense for any early prototype, but just want to check - you're also planning to add a time component in the future, right?

0_config/code/nldas_cube_utils.R
```diff
@@ -0,0 +1,57 @@
...
+
+
+calc_nldas_files <- function(time_range, time_stride, cube, ...){
+  
+  space_chunk <- sprintf("%s.%s_%s.%s", cube$x0, cube$x1, cube$y0, cube$y1)
+  variables <- c(...)
+  if (length(variables) == 0){
+    stop('must specify at least one variable as input to ... argument')
+  }
+  time_chunk_lead <- seq(time_range$t0, time_range$t1, by = time_stride)
```
aappling-usgs: these are good variable names, but this would still be easier to read with a comment here and there stating your general intention with each cluster of code lines

0_config/code/nldas_cube_utils.R
jread-usgs: I can't seem to reply to the above comment about `I()`. I was going to hold off on that and see how my small example works in practice. I worry that adding some new systems and kicking off the rebuild may cause me to forget to `delete` that target and rebuild the diff config. Anyhow, yes, I am thinking about it, but not sure right now...

jread-usgs: and for time, yes, I admit I forgot I needed a time diff. But it is needed for efficiency.

aappling-usgs: :+1:

jread-usgs: This gets more complicated, because I also need to include a _variable_ diff. Some variables may be complete for the cube, others not started, so working that up now and cleaning these other things up too. :+1:

PR #5, opened 2018-07-24
updating file structure, and result of recent loop_tasks

jread-usgs:


6_drivers_prep_rmk.yml
```diff
@@ -0,0 +1,59 @@
...
+
+sources:
+  - 6_drivers_prep/code/nldas_cube_utils.R
+  - 6_drivers_prep/code/nldas_feather_utils.R
+  
+targets:
+  6_drivers_prep: 
+    depends: 
+      - 7_nldas_subset_rmk.yml
+      - 7_nldas_append_rmk.yml
```
aappling-usgs: why `7_` instead of `6_`?

7_drivers_fetch/code/nccopy_utils.R
```diff
@@ -6,8 +6,8 @@ nccopy_nldas <- function(nc_filepath){
   message(nldas_url)
   output <- system(sprintf("nccopy -w %s %s", nldas_url, nc_filepath))
   
-  if(output){
-    unlink(nc_filename)
+  if(output | file.size(nc_filepath) == 0){
+    unlink(nc_filepath)
```
aappling-usgs: ohhh, this was the bug, huh? would have been easy to miss.

6_drivers_prep_rmk.yml
```diff
@@ -0,0 +1,59 @@
...
+  nldas_time_chunk:
+    command: c(I(5000))
+    
+  nldas_cells:
+    command: list(x=I(c(221,221,344,344,223)), y=I(c(132,196,132,196,141)), variables=I(c('dlwrfsfc','pressfc','apcpsfc','vgrd10m','ugrd10m','dswrfsfc','spfh2m','tmp2m'))) # will come from code or spatial data...
+      
+  nldas_subset_box: # this is a FULL time pull
+    command: nldas_diff_box(nldas_cells, '6_drivers_prep/in/nldas_cells.rds') 
+      
+  nldas_append_box: # this is a PARTIAL time pull, where partial time could be 0 (complete)
```
aappling-usgs: How do the full and partial time pulls work together? In the first round of data pulling? In subsequent rounds? I guess I don't really (yet) understand the meaning of "subset" vs "append" throughout this file.

6_drivers_prep_rmk.yml
```diff
@@ -0,0 +1,59 @@
...
+  nldas_subset_plan:
+    command: create_cube_task_plan(nldas_subset_files, nc_dir = I("7_drivers_fetch/out"), ind_dir=I('7_drivers_fetch/log'))
+
+  nldas_append_plan:
+    command: create_cube_task_plan(nldas_append_files, nc_dir = I("7_drivers_fetch/out"), ind_dir=I('7_drivers_fetch/log'))
+
+  "7_nldas_subset_rmk.yml":
+    command: create_cube_task_makefile(target_name, nldas_subset_plan, include = I("6_drivers_prep_rmk.yml"), packages = I(c('dplyr','ncdf4')), sources = '7_drivers_fetch/code/nccopy_utils.R')
+    
+  "7_nldas_append_rmk.yml":
```
aappling-usgs: Will 7_nldas_subset_rmk.yml and 7_nldas_append_rmk.yml get looped through in 7_drivers_fetch.yml, or within  a recipe yet to be written within this 6_drivers_prep.yml file?

I'm thinking about this coupling between 6_ and 7_. It feels sorta messy to me, but I also see why you did it (there's a lot of prep work, so a full phase devoted to prep work doesn't seem inordinate, and the fetching depends on the prep but is also pretty file-rich).

One way to proceed from here is to add the recipes that call loop_tasks to the end of this 6_ file, and to add the combination recipes to the beginning of the 7_ file. In that case I lightly question whether 6_ is really a prep rather than a fetch step, because all the fetching is really being controlled in 6_. But I also see that breaking the files there, assuming 7_drivers_fetch.yml doesn't `include` 6_drivers_prep.yml`, gives you some new/different options for ways to write and test a combination phase that doesn't require a complete build of 7_nldas_subset_rmk.yml and 7_nldas_append_rmk.yml, though you can loop back to vizstorm-GIF for some ideas on how to avoid that requirement within a single .yml, as well.

Based on my discussion about WQP with Sam yesterday, I'm also wondering whether 6_ can avoid depending on 7_ for the iterative, additive data pulling approach I think we're taking with both WQP and NLDAS. Will the task planning here use the combined results from the 7_ fetch to determine which NLDAS cells need to be pulled in the next round? If so, then maybe 6_ will need to depend on 7_? If that's the case, I might advocate a little more strongly for combining those two phases into one.

7_drivers_fetch/log/7_nldas_subset_rmk.ind
```diff
@@ -0,0 +1,9 @@
+1_nldas_cubes/log/NLDAS_time[0.4999]_y[132.196]_x[221.344]_var[apcpsfc].nc.ind: bdf7bd70c18464406179d85151e5cc36
+1_nldas_cubes/log/NLDAS_time[0.4999]_y[132.196]_x[221.344]_var[dlwrfsfc].nc.ind: ba1ae3f443eb91b10b710ad016f40dc1
+1_nldas_cubes/log/NLDAS_time[0.4999]_y[132.196]_x[221.344]_var[dswrfsfc].nc.ind: 2120fc5077f72bac272b34927f0afe18
+1_nldas_cubes/log/NLDAS_time[0.4999]_y[132.196]_x[221.344]_var[pressfc].nc.ind: bb461888348a1bfae9d31f0e13fdb59f
```
aappling-usgs: are you just calling `loop_tasks` from the R prompt for now?

6_drivers_prep/code/nldas_feather_utils.R
```diff
@@ -0,0 +1,84 @@
+create_feather_task_makefile <- function(makefile, feather_files, include, feather_dir, ind_dir, packages, sources){
```
aappling-usgs: is this function getting used somewhere?

6_drivers_prep/code/nldas_feather_utils.R
```diff
@@ -0,0 +1,84 @@
...
+    step_name = 'nccopy',
+    target = function(task_name, step_name, ...) {
+      file.path(nc_dir, task_name)
+    },
+    command = "nccopy_nldas(target_name)"
+  )
+  
+  cube_task_plan <- create_task_plan(sub_files, list(cube_task_step), final_steps='nccopy', ind_dir=ind_dir)
+  
+  create_task_makefile(
```
aappling-usgs: for looping, a reminder that you may want to separate the task makefile target from the task plan target so you can pass them both to `loop_tasks`

6_drivers_prep_rmk.yml
jread-usgs: Yes, I wrote this to make space to be able to either add new sites that expand the bounding box of the cells that are pulled or new time steps that would be added to existing files. 

"subset" is a full data pull for all time that is specified.
"append" is a data pull for only partial time (it won't pull time steps you already have). 

"append" is really rare initially, because all of the pulls are happening for the first time, but if we wanted to re-pull things next year with a longer time dimension, I think all of the tasks would end up being "append" jobs (unless we also updated the spatial domain at the same time).

7_drivers_fetch/log/7_nldas_subset_rmk.ind
jread-usgs: yes

6_drivers_prep/code/nldas_feather_utils.R
jread-usgs: Not yet, it is partial, as I am starting to work on the feather/cell data effort

6_drivers_prep_rmk.yml
jread-usgs: chewing on this...to be honest, I haven't spent much time thinking about the orchestration/linkage across the yamls, since my approach may change and I don't want to work on that part of the plumbing if it is in flux. What I am working on is the functions that build the file lists and get the data, then working on the next round of utilities that will use the raw .nc data. I think I will come back to this discussion when I have moved beyond those utilities and am happy with how they work across many files.

6_drivers_prep_rmk.yml
aappling-usgs: ahh, OK, so "subset" targets are also responsible for appending to the data, but in a spatial rather than temporal dimension?

jread-usgs: yes. I think "append" is an OK word for what those targets do, but "subset" isn't a great choice. I couldn't think of a better word and moved on and kept it for the time being

aappling-usgs: i know it adds underscores, but maybe `append_time` and `append_sites` (or `add_time` and `add_sites`)?

jread-usgs: I think I am leaning towards the following:
```yaml
  nldas_replete_box: # this is the domain that will require a FULL data pull in the time dimension
    command: nldas_diff_box(nldas_cells, '6_drivers_prep/in/nldas_cells.rds') 

  nldas_partial_box: # this is the domain that will require a PARTIAL time pull, where partial time could be 0 (complete)
    command: nldas_update_box(nldas_cells, nldas_replete_box) 
    
  nldas_partial_time:
    command: nldas_times_diff(nldas_time_range, '6_drivers_prep/in/nldas_times.rds')

  nldas_new_cube_files:
    command: calc_nldas_files(nldas_replete_box, nldas_time_range, nldas_time_chunk)
  
  nldas_append_cube_files:
    command: calc_nldas_files(nldas_partial_box, nldas_partial_time, nldas_time_chunk)
  
  nldas_new_cells:
    command: nldas_diff_cells(nldas_cells, '6_drivers_prep/in/nldas_cells.rds') 
    
  nldas_append_cells:
    command: nldas_update_cells(nldas_cells, nldas_new_cells, '6_drivers_prep/in/nldas_times.rds')
```

aappling-usgs: Cool, I like the new vs append distinction. I like the sound of replete and partial...but the replete box describes the cells where currently there is NO data, so from one perspective it seems backward, and I could almost go for `nldas_empty_box` instead...that said, in combination with the comments, I'm having no trouble understanding what you're doing now. And I can't think of a better naming scheme for the nldas_xx_box targets, so I say go for it.

PR #7, opened 2018-07-26
updates from remaining comments on earlier PR

jread-usgs:
I think I tackled the bigger remaining issues from the review of #5 
@aappling-usgs can you confirm this is along the lines of what you were thinking by having the prep and the fetch stuff together?


jread-usgs:
Yes, those cells are mostly coded, but are currently a make dead ends. The relationship between the cubes and the cells is:

image

Does that make sense? Seemed easier to draw than to type it out

aappling-usgs:
Nice! I think I had been picturing cells as being small boxes within the larger box (so 2D, or 3D but with time not a very important dimension), and cubes being cells (or larger spatial sub-boxes within the box) for specific subsets of time. Looks like cubes have larger spatial dimension than I was picturing, so that's good to know. I'm assuming time is the vertical dimension in your drawing, is that right? Do cells have just one space dimension x time, or are they actually tall skinny 3D rectangular prisms (small lat/lon widths, tall time height)?

jread-usgs:
Yes, time is the vertical dimension here. A cell is a single NLDAS grid cell, and each lake will use a single cell's data (but from multiple variables) as model input. The cell represents all of the time we need for the model to run. Because of the way NLDAS data are stored, it is much slower to get all time for a small spatial domain then to get all space for a shorter time domain, hence the need to break this up into cubes -> cells (otherwise, we'd go straight to the cells as the initial step).

jread-usgs:
exactly

PR #9, opened 2018-07-27
wqp starter + fetch crosswalk from 2017 release

aappling-usgs:
Oops, I didn't mean to include the WQP copying in this branch, but apparently I did, and guess it doesn't matter too much since I've basically just copied Sam's code from the `validation` repo straight into this one with some very slight modifications. Much more to come on that step (`6_temp_wqp_fetch`) in future PRs.

Main focus of this PR was intended to be the fetching of the shapefile, which has ID crosswalk information, from the 2017 lakes data release on ScienceBase.

I created a Google Drive folder for the project to accommodate and test that fetching, but per our conversations on Slack, in a few days we may want to switch the folder to which gd_config.yml is pointing.


aappling-usgs:
Thanks for reviewing, Sam. Yes, I'll be adding to the 6_ stuff later. And yes, I've now renamed @jread-usgs 's existing Drive folders and have repointed the gd_config.yml to that parent folder. I rebuilt the 1_crosswalk data files and can now see them in the new (old) Drive folder:

image

PR #13, opened 2018-08-08
rebuild faster feathers in the future

jread-usgs:
got some file checks and some lessons learned from the bigger builds


jread-usgs:
Alison, note this has the issue w/ log dirs and data dirs being different (.nc.ind and .nc are not in the same dirs, for example). I think it is too costly to change at this point.

6_drivers_fetch.yml
```diff
@@ -67,7 +67,9 @@ targets:
     command: loop_tasks(task_plan = nldas_append_cube_plan, task_makefile = "6_fetch_append_cubes_tasks.yml")
 
   new_feather_plan:
-    command: create_cell_task_plan(nldas_new_cells, replete_time_range, nldas_new_cube_files, feather_dir = I('6_drivers/out/feather'), ind_dir=I('6_drivers/log/feather'))
+    command: create_cell_task_plan(nldas_new_cells, replete_time_range, nldas_new_cube_files,
```
jread-usgs: oops. Noting here that `append_cubes`, `new_cubes`, and `new_feathers` should be indicator files, not objects. Not sure if changing them now would trigger a rebuild of all of the files...?

6_drivers/src/nldas_feather_utils.R
```diff
@@ -8,8 +8,9 @@ create_cell_task_plan <- function(cells, time_range, cube_files, ind_dir, feathe
     command = function(target_name, task_name, ...) {
       this_var <- parse_feather_filename(task_name, 'var') # better filtering needed?
       these_files <- cube_files[grepl(this_var, cube_files)] %>%
+        basename() %>% paste0('.ind') %>% file.path(cube_ind_dir, .) %>%
         paste("\n       \'", ., collapse = "\',", sep = "")
-      sprintf('cubes_to_cell_file(target_name, %s\')', these_files)
+      sprintf('cubes_to_cell_file(target_name, nc_dir = I(\'%s\'), %s\')', cube_nc_dir, these_files)
```
aappling-usgs: It's sorta awkward that this line has to add an orphaned single quote just before the closing parenthesis to cubes_to_cell_file. It does work, but if you wanted an alternative, you could change line 12 to
```
paste0("\n       '", ., "'", collapse = ",")
```
and then leave out that orphaned quote in this line.

6_drivers/src/nldas_feather_utils.R
```diff
@@ -99,7 +101,8 @@ cubes_to_cell_file <- function(filename, ..., nc_files = NULL){
   cell_out <- data.frame(data = rep(NA_real_, length(cell_time_indices))) %>% setNames(cell_var)
 
   if (is.null(nc_files)){
-    nc_files <- c(...)
+    nc.ind_files <- c(...)
+    nc_files <- file.path(nc_dir, tools::file_path_sans_ext(basename(nc.ind_files)))
```
aappling-usgs: you could still use `scipiper::as_data_file` in place of `tools::file_path_sans_ext` here, even though you're also swapping out the dir path - might make it slightly clearer what you're doing, or not...

6_drivers_fetch.yml
aappling-usgs: This comment is out of date with respect to new_feathers, right?

I don't think the files would get rebuilt, but I do think remake would re-check all the files represented by each of those .ind files.

.gitignore
```diff
@@ -40,6 +40,7 @@ rsconnect/
 **.nc.ind
 **_tasks.yml
 6_drivers/log/feather/*.feather.ind
+build/status/MV9ubGRhc19jdWjlcy9sb2cvTkxEQVNfM*.yml
```
aappling-usgs: does this seem to be working?

6_drivers_fetch.yml
```diff
@@ -49,10 +61,10 @@ targets:
...
   nldas_new_cube_plan:
-    command: create_cube_task_plan(nldas_new_cube_files, ind_dir=I('6_drivers/log/nc'))
+    command: create_cube_task_plan(nldas_new_cube_files, ind_dir = cube_ind_dir)
 
   nldas_append_cube_plan:
-    command: create_cube_task_plan(nldas_append_cube_files, ind_dir=I('6_drivers/log/nc'))
+    command: create_cube_task_plan(nldas_append_cube_files, ind_dir = cube_ind_dir)
 
   "6_fetch_new_cubes_tasks.yml":
     command: create_cube_task_makefile(target_name, nldas_new_cube_plan, include = I("6_drivers_fetch.yml"), packages = I(c('dplyr','ncdf4','progress')), sources = '6_drivers/src/nccopy_utils.R')
```
aappling-usgs: it seems that these two calls to `create_cube_task_makefile` have a lot of shared arguments - those could be pushed into `create_cube_task_makefile` so they'd only appear once. The consequence of making that change now should only be a rebuild of the two .yml files, at which point remake should realize that the .yml files weren't actually changed at all.

6_drivers_fetch.yml
```diff
@@ -16,19 +16,31 @@ targets:
...
   nccopy_time_stride:
     command: c(I(25000))
 
+  cube_data_dir:
+    command: c(I("6_drivers/out/nc"))
+
+  cube_ind_dir:
+    command: c(I("6_drivers/log/nc"))
+
+  feather_data_dir:
```
aappling-usgs: The other names - `cube`, `cell` - describe spatial units, but `feather` only gives the format. These `feather` files seem to be `cell` files...are there other `cell` files you're distinguishing from by using the name `feather` instead of `cell`?

.gitignore
jread-usgs: I don't think so...but I have a lot a lot a lot of these build yamls in here, and they have some pattern to them, but too many combinations to fight the battle against them this way even it it worked

aappling-usgs: Yeah, OK...someday we'll figure it out, but maybe not this go-round.

6_drivers_fetch.yml
jread-usgs: good idea.

6_drivers_fetch.yml
jread-usgs: no, you're right. This is a spot where I wasn't consistent with the naming

8_drivers_munge/src/GLM_driver_utils.R
```diff
@@ -0,0 +1,108 @@
...
+  # turn into actual driver vars:
+  drivers_out <- drivers_in %>% rename(ShortWave = dswrfsfc, LongWave = dlwrfsfc) %>%
+    mutate(WindSpeed = sqrt(ugrd10m^2 + vgrd10m^2), AirTemp = tmp2m - 273.15,
+           RelHum = 100*spfh2m/qsat(AirTemp, pressfc*0.01),
+           Rain = apcpsfc*24/1000) %>%
+    mutate(Snow = ifelse(AirTemp < 0, Rain*10, 0), Rain = ifelse(AirTemp < 0, 0, Rain)) %>% #convert to m/day rate)
+    select(time, ShortWave, LongWave, AirTemp, RelHum, WindSpeed, Rain, Snow)
+
+  # now downsample?
+  readr::write_csv()
```
jread-usgs: this is unfinished and snuck in w/ my other PR review updates. it is close, though. Creates a proper hourly driver data.frame, but I didn't start writing them to file because I assume I'll want to downsample to daily. Only question was whether I'd have an argument in this function to allow us to use it to write daily vs hourly files.

aappling-usgs:
Ready for merge?

PR #15, opened 2018-08-10
ready for the all lakes build

jread-usgs:


jread-usgs:
Another open question that I am a little worried about: When I update `nldas_cells` with the contents of `nldas_cells_COMPUTED`, we're going to have a task makefile with ~16K targets. We'll see what performance looks like...

jread-usgs:
Well, here goes nothing:
image

☝️ all modeled lakes within loop_tasks

jread-usgs:
Sure! I will have a new PR up later w/ the alternative build (one-to-many) pattern

PR #17, opened 2018-08-14
One-to-many targets for feather files

jread-usgs:
This utilizes the scipiper `sc_indicate` function updates that allow us to hash many files in the `data_file` input. 

The `append` cells part of this isn't yet updated, but I won't have time (or a use-case) to tackle that in the near future. All of the driver data are built for the FY18 lakes now w/ this PR.


aappling-usgs:
Cool, FY18 ready to go! Which of the .ind files are going to be depended on (directly or indirectly) by downstream steps? For those we will need the corresponding build/status file so that Sam's and my computers don't try to rebuild them.

jread-usgs:
@aappling-usgs I'm not yet sure what the downstream depends are on these - the only stage that is reliant on the drivers is the Yeti/Condor workflow, which is probably outside of the pipeline. Do you have downstream targets that should depend on any of these? The ones that come to mind are the "lake to cell-driver" crosswalk table, which isn't a target yet.

aappling-usgs:
Yeti/Condor (for running GLM) is Pipeline 2, right? So yeah, we don't need the .ind or build/status files for those operations.

I don't have any downstream targets yet, but I assume there will be that "lake to cell-driver" crosswalk table and also whatever function packages up the inputs for each lake. What info will those crosswalk & packaging functions need to have to be assured that the NLDAS data are available on the current computer?

The .ind files in this commit are:
1. 6_drivers/log/feather/6_fetch_new_cell_tasks.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[apcpsfc].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[dlwrfsfc].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[dswrfsfc].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[pressfc].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[spfh2m].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[tmp2m].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[ugrd10m].yml.ind
1. 6_drivers/log/feather/cellgroup_time[0.346848]_var[vgrd10m].yml.ind
1. 8_drivers_munge/out/8_drivers_munge_tasks.ind

I think it makes sense for any .ind file worthy of committing to also have a build/status file committed...I guess unless the .ind file is _only_ useful to other humans as a readable record of what was done or when. If anybody else's machine might use an .ind file, or might try to, then both .ind and build/status files should be shared. But now that I've dug through your PR for a while...I guess maybe items 1-9 above might fall into that first category, since they're things that only your computer will run but it might be nice for everyone else to know?

I'm not sure about 10. It raises the question of how the packaged inputs get shared with others and/or Yeti/Condor. Is that something you plan to script, or will you just be using `scp` or external hard drive to transfer the data? Relatedly, do we plan to package the outputs (temperature data) along with the inputs (NLDAS), or will they stay separate until Pipeline 3? If either of those things will be remake-scripted, we may need you to commit the .ind and build/status files for 10 and its dependencies going back up the chain within Pipeline 1.  Or we could choose to break the remake pipeline with `I()` or unmentioned dependencies to get those latter tasks done from your computer. What are you picturing?

Side note: I thought we were using 7_drivers_munge for munging, to leave room for 8_package_inputs and 9_reality_check? https://docs.google.com/document/d/1iwwhVUIjcxOFy1GT5i9x_yLAYJbldLFOYuGVpZxS8Vw/edit#

jread-usgs:
Ok, I understand. I need to pair the .inds w/ the build yamls. Trouble is, I have hundreds (thousand?) of build yamls that aren't checked into this repo. I'd need to parse those to figure out which ones need to be checked in. 

For questions about 10 - I hadn't thought about the needs for pipeline 3 yet, and had assumed I'd be doing a `scp` to get the files to Yeti. The actual driver files are much smaller than the raw data, since they are only daily (instead of hourly). 

I'd welcome feedback on how this connects to pipeline 3. If, for example, there are only 10 lakes modeled w/ PGDS, I would probably advocate for a secondary build for those specific lakes and the sharing of those files/inds using the cache. But, I think we can do better than 10, so that might warrant a different approach.

jread-usgs:
Whoops, will change to 7_

aappling-usgs:
OK, if you'll be using `scp` and 8_drivers_munge is a dead end with respect to Pipeline 1 (no merging with other data, or merging the crosswalk with other crosswalks), then I think all of your .ind files are optional and it doesn't matter whether you commit them (or their build/status files) or not.

I know the build/status gitignore thing is painful right now. Is it too late to consider implementing https://github.com/USGS-R/scipiper/issues/51 for this project? For just 10 files, though, you could git add them manually with the help of RStudio Ctrl-Shift-F or a small script. The build file corresponding to 6_drivers/log/feather/cellgroup_time[0.346848]_var[dswrfsfc].yml.ind is the one that contains the text `name: 6_drivers/log/feather/cellgroup_time[0.346848]_var[dswrfsfc].yml.ind`, etc. So you could find those files and then manually commit them. To find using R:
```r
buildfiles <- dir('build/status', full.names=TRUE)

build2ind <- sapply(setNames(nm=buildfiles), function(buildfile) { yaml::yaml.load_file(buildfile)$name })

names(build2ind)[match(c('6_drivers/log/feather/6_fetch_new_cell_tasks.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[apcpsfc].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[dlwrfsfc].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[dswrfsfc].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[pressfc].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[spfh2m].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[tmp2m].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[ugrd10m].yml.ind',
'6_drivers/log/feather/cellgroup_time[0.346848]_var[vgrd10m].yml.ind',
'8_drivers_munge/out/8_drivers_munge_tasks.ind'), build2ind)]
```

From there you could exclude those 10 files from a general .gitignore rule for `build/status` using the `!` notation in .gitignore.

...or, in this case, sounds like you could treat those 10 files as optional FYI files and never mind about the build/status.

jread-usgs:
Thanks for the feedback - I used that info to clean up my build files (including deleting ~thousand that pointed to old files that don't exist anymore). Now it is much more manageable. The only files that are in limbo are build files that correspond to the .ind.nc targets, to which there are 112. 

I'm not sure how to make https://github.com/USGS-R/scipiper/issues/51 happen for this project, and I have the sense that making that change would trigger a rebuild of the most expensive targets, the raw .nc data. Is that not the case? Dealing with the 112 build yaml files that are in limbo right now isn't a pain. it was starting to bog down having thousand(s) in there, but this is workable for me for now.

jread-usgs:
Thanks - great points and good questions. 

I like the suggestion to make this switch in early FY19. FY19's work includes 30K lakes total, so let's get it right!

But to answer your other question (and I'd be up to rebuild in 18 if this helps), it took 5 days to pull all of the .nc files. I think that could be significantly faster if we implemented that parallel `loop_tasks` solution. The remainder of the tasks (feather cells, and driver files) take hours instead of days after the work in this `one2many` PR.

aappling-usgs:
I knew it took a long time - good to get a number for it. 5 days is a lot! Sounds good to aim for early FY19.

7_drivers_munge.yml
```diff
@@ -0,0 +1,39 @@
...
+targets:
+  7_drivers_munge:
+    depends:
+      - "7_drivers_munge/out/7_drivers_munge_tasks.ind"
+
+  meteo_file_dir:
+    command: c(I("7_drivers_munge/out"))
+
+  cell_group_files:
+    command: calc_cell_group_files(nldas_cells, replete_time_range, ind_dir = cell_ind_dir) # where is the side-effect of writing to '6_drivers/in/nldas_times.rds'?? (updates cubes)
```
aappling-usgs: Is this comment/question still valid, and if so, is it something you want to discuss, and if so, could you elaborate?

6_drivers/src/nldas_feather_utils.R
```diff
@@ -1,29 +1,46 @@
-create_cell_task_plan <- function(cells, time_range, cube_files, cell_data_dir, cell_ind_dir, cube_data_dir, cube_ind_dir){
+create_new_cell_task_plan <- function(cells, time_range, cube_files, cell_data_dir, cell_ind_dir, cube_data_dir, cube_ind_dir){
+
+  # from https://stackoverflow.com/questions/17256834/getting-the-arguments-of-a-parent-function-in-r-with-names
+  # we want to get the name of the object passed to `cells`, so we can refer to it elsewhere
+  cl <- sys.call(0)
+  f <- get(as.character(cl[[1]]), mode="function", sys.frame(-1))
+  cl <- match.call(definition=f, call=cl)
+  cell_obj <- as.list(cl)[-1][['cells']] %>% as.character()
```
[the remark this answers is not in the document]
jread-usgs: Looks much better. I think the example I stole for this was reaching back one further function call (like the parent of this one), so I used +1 on the indices of the example, and used the syntax you see above. In this example though, I don't need to do that and `substitute` is indeed much cleaner.

7_drivers_munge.yml
aappling-usgs: OK, happy to discuss this in FY19.

PR #23, opened 2018-09-19
adding readme for file dependencies in pipelines #2 and #3

jzwart:
See the conversation [here ](https://github.com/USGS-R/lake-temperature-process-models/pull/2#discussion_r218229836) about how to go about dealing with file dependencies from other repositories.


jzwart:
We could eventually try automating the sharing / pushing / pulling data files and indicator files using `scipiper`, but as for now, we'll stick to manually pushing these to respective folders

aappling-usgs:
I like the goal. The google drive piece would be pretty straightforward, I think, but I'm less sure how we'll manage the .ind files - do we have to assume that the user has the four pipeline repos all local, and all in sibling folders with the same parent? Or is there a way to git commit straight to GitHub?

jzwart:
@limnoliver and I were just talking about this and also weren't sure how to deal with the indicator files, and were hoping that you, @aappling-usgs, would have some answers :) 

We were thinking that maybe the dependent pipeline could also pull the indicator file from github repo, but this would mean that the .ind file _always_ gets pulled down but not necessarily the data files. Then you wouldn't have to make sure that everyone has all 4 repos in the same sibling structure

aappling-usgs:
Wish I had more answers =). Interesting idea to have the dependent pipeline pull the .ind file - that does sound easier than figuring out how to create a git commit and/or PR to another repo from within R! There's the usual problem that you can't set remake up to rerun anything whose local inputs are unchanged, except by using `force=TRUE`. But (1) in a way that's nice, because someone in the middle of something on pipeline2 might not be prepared to adopt a change from pipeline1 right away, and (2) if we don't like it, one solution to consider is adding an .Rprofile file to everybody's repo that either (2a) goes directly to github and pulls down the .ind file if changed, or (2b) edits some dummy input to the remake recipe for retrieving that .ind file. Either 2a or 2b would effectively cause a force=TRUE once per session -- maybe that would be an acceptable frequency?

jzwart:
Ah yeah that's a good idea! I think the the once per session would be frequent enough - I think the dependent data files will be the lakes themselves (e.g. geospatial data) and the observations. The features (lakes) likely won't change very much and the observations probably (?) won't get updated very frequently..

Issue #27, opened 2018-11-28
add local time offset for daily driver summaries


jread-usgs:
Currently, the daily driver data are assembled as averages of the values between midnight and the next midnight, UTC. We shouldn't use UTC for the midnight cutoff, and use a local tz or something relevant to solar noon so we don't blend data from two daylight periods into one.

aappling-usgs:
Relevant code line that will need to be adjusted: https://github.com/USGS-R/lake-temperature-model-prep/pull/29#discussion-diff-259084172R144

PR #29, opened 2019-02-06
Local drivers and updates for WRR paper

jread-usgs:
Increased the time range and local meteo builders


6_fetch_local_meteo.yml
```diff
@@ -0,0 +1,101 @@
...
+sources:
+  - 6_drivers/src/local_meteo_utils.R
+  - 7_drivers_munge/src/GLM_driver_utils.R
+
+targets:
+  6_fetch_local_meteo:
+    depends:
+      - 7_drivers_munge/mendota_out/mendota_local_meteo.csv
+
+
```
aappling-usgs: Are we likely to bring in local data for many more lakes?  I think the answer is no, in which case we might as well just merge this and let it coexist with the main many-lakes pipeline. But if the answer is yes, then `aoss` is cryptic and `airport` is quite generic, so I'd advocate for using the lakes IDs to name these functions and targets.

7_drivers_munge/src/GLM_driver_utils.R
```diff
@@ -141,7 +141,7 @@ feathers_to_driver_file <- function(filepath, cell_group_table){
            Rain = apcpsfc*24/1000) %>%
     mutate(Snow = ifelse(AirTemp < 0, Rain*10, 0), Rain = ifelse(AirTemp < 0, 0, Rain)) %>% #convert to m/day rate)
     select(time, ShortWave, LongWave, AirTemp, RelHum, WindSpeed, Rain, Snow) %>% # now downsample?
-    mutate(date = lubridate::as_date(time)) %>% group_by(date) %>%
+    mutate(date = lubridate::as_date(time, tz = "Etc/GMT+6")) %>% group_by(date) %>%
```
aappling-usgs: Is it always appropriate to specify GMT+6? Would this work on my computer in EST? What was the motivation for this change?

6_drivers/in/AOSS_solar_README.md
```diff
@@ -0,0 +1,389 @@
+**source
```
aappling-usgs: This looks like more a .txt than a .md file. Doesn't really matter, but the .md extension is causing GitHub to display the whole thing in bold font, and it's possible there are some other formatting surprises in here...

6_drivers/in/AOSS_solar_README.md
```diff
@@ -0,0 +1,389 @@
...
+These instruments are sensitive to the same broadband spectral range as the NIP 
+and PSP, 280 to 3000 nm.  They are also calibrated on a biennial basis at 
+the NOAA calibration facility in Boulder, CO
+
+5.      Precision Infrared Radiometer (PIR) 
+
+The PIR measures and downwelling thermal infrared irradiance at Madison only.
+It is sensitive to the spectral range from 3000 to 50,000 nm.  NOAA maintains
+three standard PIRs that are calibrated annually by a world-reputable
+organization.  These standards are used to calibrate field PIRs.
```
aappling-usgs: The file 6_drivers/in/ntl129_2_v5.csv is too big to view in GitHub. Is the repo the best place to be storing this file, or should it be moved to Drive?

7_drivers_munge/src/GLM_driver_utils.R
jread-usgs: No, this is a footprint-specific change, making sure the hourly data end up summarized to the proper date. Two of the three states are in CT, so that is why this is used, and only lakes from those two states were in the WRR analysis.

6_fetch_local_meteo.yml
jread-usgs: I think the answer is no - I think this is paper-specific and not something we should use generally. I'm wondering if I should've PR'd this into a branch that stays solo, as it is related to getting things together for the paper, not necessarily for the project itself.

6_drivers/in/AOSS_solar_README.md
jread-usgs: Yes, it should either be on drive or we should make a call to their server for the file w/ a fetch.

6_fetch_local_meteo.yml
aappling-usgs: Maybe a separate branch would be best, so this gets preserved somewhere we can find it again and where we're unlikely to accidentally delete it in next year's spring cleaning. But also, some messiness with multiple products from a project seems to be inevitable based on my experience to date, so maybe we just accept it.

7_drivers_munge/src/GLM_driver_utils.R
aappling-usgs: Hmm, so maybe now would be the time to make this an if-else or switch statement based on the state? or at least make an issue for it - this is the kind of thing I'll forget now and never remember.

jread-usgs: Yes, there is an issue for it: https://github.com/USGS-R/lake-temperature-model-prep/issues/27

6_drivers/in/AOSS_solar_README.md
jread-usgs: fixed

6_fetch_local_meteo.yml
jread-usgs: I am going to kill this an open up a PR onto a branch here.

7_drivers_munge/src/GLM_driver_utils.R
aappling-usgs: Oh, excellent!

Issue #31, opened 2019-02-08
restructure 7a_temp_coop_munge to avoid individual builds when cooperator files added


limnoliver:
Currently, parsed cooperator files are ignored, and only the combined munged data is pushed to google drive. This means that if a new cooperator file is added and someone else (other than the original parser computer) wants to update, all files have to be downloaded and parsed.

Options might be to:

1. Push the individual munged files to google drive
2. Have new files simply added to the old combined file, assuming that all old data are "good" and we are simply missing new files.

aappling-usgs:
How long does it take to download and munge all the files?

limnoliver:
Not too long (~2 hours?). The biggest hurdle seems to be OS-specific issues with accessing .accdb files that trips the process up. So then, a third solution would be to fix that issue, and having local rebuilds wouldn't be too big of an issue.

aappling-usgs:
Hmm, gotchya. Or always run this phase on a computer with the right OS (and ideally also with some cooperator files already processed)?

limnoliver:
Yeah exactly. Currently that's me.

limnoliver:
I think I've come around to the opinion that we should leave this part of the pipeline as-is, but make some changes on the next rebuild to add more defenses against rebuilds and failed rebuilds. These include:

1) Saving access database files in a different format on Google Drive (outside of the repo) so that we don't have to deal with Access-to-R issues that seem to be computer-specific.
2) Sort file names when creating `coop_all_files`
3) Write and indicate a file for `coop_all_files` instead of an R object. Currently, all users have to force rebuild this step.

jread-usgs:
We have a potential task related to this in the document generated to onboard @lindsayplatt 

Is that higher priority and more well-defined now?

limnoliver:
Yes, this is a pretty discrete task now, and guessing it would take me an ~hour. I think it can wait until the next time we're in mucking around in the coop files (e.g., we decide to parse more files) since it will trigger a rebuild.

lindsayplatt:
I believe all 3 of the issues noted in [this comment above](https://github.com/USGS-R/lake-temperature-model-prep/issues/31#issuecomment-574729935) were addressed in #146

PR #32, opened 2019-02-08
Finish incorporating new cooperator data

limnoliver:
Having Jordan try to run this brought up new issues of computer-specific issues with accessing .accdb files and also with individuals having to rebuild munged cooperator files. Need to reconsider this workflow, as documented [here](https://github.com/USGS-R/lake-temperature-model-prep/issues/31).


aappling-usgs:
This PR has been sitting for about 2 weeks now. Would it be OK to merge and assume that #31 is a task for a different PR? @jread-usgs @limnoliver

limnoliver:
Yes @aappling-usgs - we can address that later. In the meantime, the most efficient way to rebuild if there is new cooperator data is to have me do it.

PR #34, opened 2019-02-22
Local drivers and other file experiments for WRR initial submission

jread-usgs:


jread-usgs:
These files are being created on my computer, and then the drivers are synched to Yeti manually when we are ready to run. 

I think it would be best for the smaller files (everything but the NLDAS drivers) to be on drive, but I think I've backed myself into an odd corner on that, because while this repo code holds the different branches, the google drive wouldn't, and therefore probably should've been a different project.

aappling-usgs:
OK. Yes, it's complicated, but it sounds like we're probably not corrupting the build by merging this as its own branch.

Issue #48, opened 2019-05-01
new footprint crosswalks not properly using shared cache


limnoliver:
Not exactly sure what the solution is (yet), but I could not build `wqp_want_lakes` as it kept getting hung up on (already in the shared cache) `1_crosswalk_fetch/out/winslow_lakes_sf.rds.ind`. This is a dependency issue that was solved with a workaround by running 

```
gd_get('1_crosswalk_fetch/out/winslow_lakes_sf.rds.ind')
gd_get('1_crosswalk_fetch/out/lakes_shapefile.prj.ind')
gd_get('1_crosswalk_fetch/out/lakes_shapefile.shx.ind')
gd_get('1_crosswalk_fetch/out/lakes_shapefile.dbf.ind')
```

So it seems the current pipeline does not trigger `gd_get` of the other geo files that need to come along with the shape file. I think I may have been able to just do the last 4 commands, but not sure now. Maybe the easiest solution would be to add the above file to the `depends` field for `1_crosswalk_fetch`.

jread-usgs:
Yes, this becomes easier if we never rely on the individual shapefile files. Instead using the .zip as the target is one way around it too.

jread-usgs:
We no longer rely on these individual files. I think this issue no longer applies.

PR #51, opened 2019-05-01
updated WQP pull to expanded footprint

limnoliver:
@jread-usgs - related to #48, some of the build status files for the crosswalks were updated when I built the WQP pull since the winslow indicator file did not exist, so everything downstream of that was updated. I think this is okay but might make driver steps out of date? The files themselves aren't different so not exactly sure what this would trigger downstream. See the [indicator ](https://github.com/USGS-R/lake-temperature-model-prep/commit/0348e9d523082876278d762a8ad83b42ddf0629b) and [build ](https://github.com/USGS-R/lake-temperature-model-prep/commit/aae54179ebe08b039472b552fc439a2cff2073fc) commits.


jread-usgs:
This should be ok - as long as the ind for the `2_crosswalk_munge/out/lakes_sf.rds` file didn't change, the footprint for the drivers won't change. 

But I see that it did get updated, so the question is whether the change is meaningful. For it to mean something for the drivers, the bbox of the `2_crosswalk_munge/out/lakes_sf.rds` would need to be different...but I guess we can't tell if that is the case w/o you checking yours and me checking my local version. 

I have
```
xmin      ymin      xmax      ymax 
-99.72474  40.91053 -83.05733  49.00011 
```

What do you have?

limnoliver:
I have: 
```
     xmin      ymin      xmax      ymax 
-99.72474  40.91053 -83.05733  49.00011 
```
so I think we're good?

jread-usgs:
I think were good, yes. 🕺

PR #64, opened 2019-07-09
Add MN bathy to summary table and map

vitensek:


8_viz.yml
```diff
@@ -27,7 +27,8 @@ targets:
       temp_data_ind = '7b_temp_merge/out/merged_temp_data_daily.feather.ind',
       lake_names_ind = '2_crosswalk_munge/out/lake_names_crosswalk.rds.ind',
       dow_cross_ind = '1_crosswalk_fetch/out/nhd2dowlknum.RData.ind',
-      state_cross_ind = '4_params_munge/out/nhd2states.rds.ind')
+      state_cross_ind = '4_params_munge/out/nhd2states.rds.ind',
+      MN_bathy_ind = '4_params_munge/out/MN_bathy.rds.ind')
 
   8_viz/inout/lakes_summary.feather:
     command: gd_get('8_viz/inout/lakes_summary.feather.ind')
```
jread-usgs: Can't comment on the file, so doing it here. If we are using an .ind, we shouldn't be checking the map into the repo (should gitignore it). For 8_viz/out/lakes_summary_fig.html

limnoliver: Oops, this was my bad! Kelsey, can you add `html` to the `gitignore`?

jread-usgs: Need to do one more step too: https://stackoverflow.com/questions/1274057/how-to-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore

vitensek: I think I took care of it -- let me know if there's anything else!

jread-usgs: Looks good. I think sam is on vacation but when she's back, I will defer to her for :+1: :-1: on merge

build/status/N2FfdGVtcF93cXBfbXVuZ2Uvb3V0L2FsbF93cXBfZGF0X2xpbmtlZC5mZWF0aGVyLmluZA.yml
```diff
@@ -2,9 +2,9 @@ version: 0.3.0
 name: 7a_temp_wqp_munge/out/all_wqp_dat_linked.feather.ind
 type: file
 hash: 52efe208394fc69014a436993c482131
-time: 2019-05-09 14:28:22 UTC
+time: 2019-07-03 20:13:59 UTC
 depends:
-  7a_temp_wqp_munge/tmp/all_wqp_dat.rds.ind: 7fe951654e7d640d6487863cbc78f5b5
+  7a_temp_wqp_munge/tmp/all_wqp_dat.rds.ind: 2b24448d789fc14cd8bc78f9f60ae842
```
limnoliver: I'm having a hard time understanding why the WQP data was rebuilt. This indicates that the `all_wqp_dat.rds.ind` has a new hash, but there's no new build status file for that target. Is there a modified build status file for `7a_temp_wqp_munge/tmp/all_wqp_dat.rds.ind` that needs to be committed and pushed?

vitensek: Hmm...there's nothing left to commit and push on my end. Is there something I can do to investigate or resolve the issue?

jread-usgs: perhaps rebuilding the individual target `scmake(..., force = TRUE)`?

vitensek: I may have forced a rebuild of this target when I was trying to figure out why I was getting an error when I tried to build 8_viz as a first step -- it turned out to be because the test parser I wrote didn't save lakeid as a character vector, which caused a problem at the temp merge stage. But it took me a bit to figure out where the problem was, and I likely forced a rebuild here to rule this piece out. What should I do differently next time?

jread-usgs: It should be fine. If this hash is different but the build file didn't change, perhaps someone else has an uncommitted build file for this? Either way, should be fine other than causing an unnecessary rebuild. @limnoliver you see other issues?

limnoliver: Yep - should be fine.

PR #86, opened 2019-10-23
updates for new bathymetry and max depth sources for WI lakes

jread-usgs:
Now a total of 13,368 lakes w/ depth info


jread-usgs:
This is pretty vanilla. Merging myself.

PR #87, opened 2019-10-23
Added indicator file for mndow percentage hypsos zip file

vitensek:
@jread-usgs 
Note that these csv's have the same naming convention as those in '3_params_fetch/in/hypsos_m' so some of the code in munge_mndow_bathy() may be useful for pulling out the mndow id.


jread-usgs:
Kelsey - I am replacing the other hypos with a zip of the same hypos too (and using the .ind as you have here). The previous code was pulling each csv individually, and wasn't able to detect changes on drive, so it wouldn't know if there were new or modified files. Thanks for doing this. I will have my processing code up later.

PR #88, opened 2019-10-24
New pattern for bathy/hypso files, now merging bathys when multiple

jread-usgs:


jread-usgs:
Merging this one to get things prepped for @limnoliver 's work on coop files. 

I think I've got WQP temperature wrapped up here, although I didn't plug things into `7b_temp_merge`, as I think that happens after coop files. 

I did check to see how many NHDHR lakes we have covered with temperature data from WQP after this pull - the count is 5,279, and 4,240 of these currently have lake depths (i.e., we can model them, shown in blue). Of the 1,039 lakes we lose (red) because of not having depth, those had 432,380 temperature records, which is a pretty big chunk (total records before joining w/ depth is 1,493,790). But, looks like most of those are great lakes data

test_lakes

I've modeled 800 of these w/ the new GLM3, and that has worked well as far as I can tell, and none of those lakes failed for reasons I couldn't explain and later fix (usually issues w/ bathymetry). I haven't compared any of the modeled temperatures with observed, or done any kind of comparison between GLM2 uncalibrated vs GLM3, since our previous work w/ GLM2 used different lake IDs and we don't have a crosswalk between Winslow et al. and NHDHR (yet).

jread-usgs:
We're surprisingly light on Vilas county data ☝️ 
I know we don't have an LTER pull (or Birge/Judy dataset), but surprised more of those lakes don't have data in WQP

PR #89, opened 2019-10-27
Some clean up before temperature merge w/ coop

jread-usgs:


jread-usgs:
Merging my own as this is just renaming and rebuilding

PR #92, opened 2019-10-29
a couple more hypsos from Kelsey

jread-usgs:


jread-usgs:
small update. merging

PR #101, opened 2019-11-12
Update coop data

limnoliver:
Updated temperature observation portion of the pipeline to accommodate NHDHR changes. 

Also updated the way the data are reduced down to daily values. It's mostly the same approach (though, simpler code), but has a few "smarter" ways like selecting the most representative sites if we are unsure if the data are from multiple sites or multiple times, and selecting the "best" source-site when multiple sources or sites are available at a given lake-date-depth.

@jread-usgs - the last thing left to do for data release is to clean up the `source` column, and create the companion `source metadata` file. Will do that once Gretchen et al provide metadata.


7a_temp_coop_munge/src/crosswalk_fxns.R
```diff
@@ -38,16 +39,20 @@ crosswalk_coop_dat <- function(outind = target_name,
     left_join(wbics_in_dat)
 
   # dow
+
+  dat_filt_dow <- filter(dat, !is.na(DOW)) %>%
+    filter(!DOW == '') %>%
+    mutate(DOW = ifelse(nchar(DOW)==8, DOW, paste0('0', DOW)))
```
jread-usgs: verified that there are only 8 char DOWs in the crosswalk (12721 of them)

7a_wqp_munge/src/munge_wqp_dat.R
```diff
@@ -67,28 +67,18 @@ munge_wqp_secchi <- function(outind, wqp_ind){
   gd_put(outind, outfile)
 }
 
-crosswalk_wqp_dat <- function(outind, wqp_munged, wqp_crosswalk, wqp_latlong_ind) {
+crosswalk_wqp_dat <- function(outind, wqp_munged, wqp_crosswalk) {
```
jread-usgs: not worth changing, but I have been trying to get in the habit of making these arg names align w/ the type. So when I see `wqp_munged`, I think it is an R object, vs `wqp_munged_ind` I would know is a file. Not a big deal at all, but something we might want to consider as we continue to evolve our collaborative pipelines - how much standardization do we do?

limnoliver: I like that @jread-usgs -- I've found the standardization of organization and file names to be super helpful when developing, so I'm happy to implement that pattern.

PR #108, opened 2019-11-15
update after running Kelsey's code

jread-usgs:


jread-usgs:
@vitensek this is the result of your changes

PR #112, opened 2019-11-27
Rebuild coop dat +

limnoliver:
This was originally meant to rebuild the coop data after Jordan added a smarter way to track changes to the parser files.

But, ultimately, in addition to the rebuild, this PR:

1. Adds the Ten Mile Lake cooperator data and parsers
2. Downloads the source metadata that has been populated, and translates our "source" tracking from file name to a unique source identifier. A source metadata file is then written that summarizes all data sources, with associated description and collection information. 

So, for the downstream activities:
- the daily data for modeling can be found in either the `7b_temp_merge/out/merged_temp_data_daily.feather` or `7b_temp_merge/out/temp_data_with_sources.feather`. These are the same data, but with different source identifiers. The first file has a more direct connection to which file the data came from. The second file has a more high-level tracking of the data source. 
- The data release should consist of a filtered version of `7b_temp_merge/out/temp_data_with_sources.feather`. Filter out sources that have a `no` in the `public_release` column of `7b_temp_merge/out/source_metadata_for_release.csv`. The source metadata file (in full) should accompany the data release. N values from the source_metadata report the number of observations (not unique dates) from each source.


limnoliver:
Some tallies as a reference:

- 89 lakes with 200 dates with 5+ depths (WRR rules)
- 866 lakes with 50+ dates
- 405 lakes with 100+ dates
- 193 lakes with 200+ dates

PR #113, opened 2020-01-03
gnisname_nhdhr_xwalk was missing build/status yml

lindsayplatt:
The build/status yml file for `gnisname_nhdhr_xwalk` was missing from the repo. So, when I went to build this repo locally for the first time, it was rebuilding this step. 

This commit has two small things that I had changed before when I was first building the repo (line ending + comment about a library).


jread-usgs:
Can you remove the changes to the two remake yaml files?

aappling-usgs:
Is the change to the contents of gnisname_nhdhr_xwalk expected, @jread-usgs ? (as indicated by the different hash in the .ind file)

jread-usgs:
At this point we are unsure. Lindsay will post some questions/comments in `ds-pipeline` to point out what we discovered about status files and force rebuilding the `_promise`. I am assuming that issue is related to this rebuild, but unclear at this point

PR #117, opened 2020-01-15
Update time for secchi

jread-usgs:


aappling-usgs:
Curious: how long did this update take?

jread-usgs:
5.5 hours

PR #119, opened 2020-01-15
Plot lake observation counts

aappling-usgs:
Pairs with https://github.com/USGS-R/lake-temperature-process-models/pull/16.

Pulls some code from lake-temperature-process-models to figure out which lakes have all the necessary parameters to run GLM, then uses that info along with temperature observation counts to make the following plot:

lake_obs_counts


limnoliver:
I like the plot, Alison! Cool to see how many more lakes have lots of dates (1000+) in the 1 depth category compared to the 5 depths category.

PR #120, opened 2020-01-16
remove triangle because uninformative

aappling-usgs:
revision based on convo with jordan just now


8_viz/src/plot_lake_obs_counts.R
```diff
@@ -31,20 +31,15 @@ plot_lake_obs_counts <- function(
     site_counts_mat <- matrix(
       NA, nrow=length(mins_obs_per_date), ncol=length(mins_obs_dates),
       dimnames=list(min_obs_per_date=mins_obs_per_date, min_obs_dates=mins_obs_dates))
-    for(min_obs_per_date in mins_obs_per_date) {
+    site_counts_mat['0', '0'] <- obs_counts %>% pull(site_id) %>% unique() %>% length()
```
aappling-usgs: I revised how I calculated the lone triangle and circle just to make sure I got it right - I think this code is more straightforward, but it produces the same numbers as before

8_viz/src/plot_lake_obs_counts.R
```diff
@@ -67,13 +62,13 @@ plot_lake_obs_counts <- function(
 
   ggplot(data=site_counts, aes(x=obs_dates, y=n_sites, color=obs_per_date)) +
     geom_line(data=filter(site_counts, obs_dates > 0), aes(linetype=has_glm)) +
-    geom_point(data=filter(site_counts, obs_dates == 0), aes(shape=has_glm)) +
+    geom_point(data=filter(site_counts, obs_dates == 0) %>% mutate(n_sites = ifelse(has_glm == 'Yes only', n_sites, NA)), aes(shape=has_glm)) +
```
aappling-usgs: removing the triangle

jread-usgs:
Looks good

PR #121, opened 2020-01-19
workaround WQP fix

jread-usgs:
There we some lakes missing in the calibration run that used to be there. I dug into that and found that one of the secchi partitions was an empty data.frame and shouldn't be...then ran into the WQP service issues that are documented elsewhere (see [here](https://github.com/NWQMC/WQP-WQX-Services/issues/263)) and then finally landed on this workaround.

After seeing that the major issues with the WQP services is just the site query's geojson return (which we use for `resultCount`), I just punted and used the csv/tsv output instead and set all resultCounts to 200. This gave me a lot more partitions, but was a viable workaround for this issue. I think this PR shows that we are back to where we should've been with secchi after adding back the sites that were in that lost partition originally...


jread-usgs:
was hoping to get a visual diff on the lake counts fig, but no luck. It has a slight increase in the solid lines on the left side. No change to dashed line (as far as I could see).

jread-usgs:
merging my own here - results are already in cache

PR #130, opened 2020-02-27
Add to mtg analysis w/ MN-specific filters

lindsayplatt:
Still not pipelined, but has some new filtering & some basic maps created during our discussion.


lindsayplatt:
Now coded into pipeline (not built), but not sure what to do with the "in" files that Kelsey shared via Google Drive.

aappling-usgs:
> Now coded into pipeline (not built), but not sure what to do with the "in" files that Kelsey shared via Google Drive.

I haven't looked at your specific case, but in general one solution for input files provided by collaborators is to manually place the file in the 'in' folder on Drive, then use scipiper functions (I actually forget which, sorry!) to create an indicator file in the repo, then have a gd_get command to get the file. There must be an example (or several) of this scattered through our various pipelines...

limnoliver:
It's `sc_indicate` -- though there's a trick to getting the options right in the function to get the indicator file you're looking for.

jread-usgs:
I thought in this case you use `gd_confirm_posted`?

aappling-usgs:
Yeah, that's the one!

8_viz.yml
```diff
@@ -70,3 +71,34 @@ targets:
...
   8_viz/out/lakes_summary_fig.html:
     command: gd_get('8_viz/out/lakes_summary_fig.html.ind')
+
+  ##### Adding summary and visual of data that we are thinking of using for TOHA
+
+  # These datasets help visualize and summarize the data we are using for TOHA models.
+  # They were downloaded from here: https://drive.google.com/drive/u/4/folders/13w3QDXicjHVymVW1NeyJpJqVRZ7KvlTD
+  # Then uploaded to this pipeline GD folder & then incorporated into the pipeline using `gd_confirm_posted`.
+  8_viz/in/WAE_numYears_meanCPUE.csv:
+    command: gd_get(I("8_viz/in/WAE_numYears_meanCPUE.csv.ind"))
```
jread-usgs: why does it need to be wrapped in `I()` here?

lindsayplatt: That would be copy-paste spaghetti. Hold on a sec

jread-usgs:
Looks good now, but we'll also need to make sure the `Kw_file` information gets in to the pipeline properly. Probably not something that needs to be taken care of on this PR, but can you create an issue for it?

lindsayplatt:
> Looks good now, but we'll also need to make sure the `Kw_file` information gets in to the pipeline properly. Probably not something that needs to be taken care of on this PR, but can you create an issue for it?

Issue created: https://github.com/USGS-R/lake-temperature-model-prep/issues/135

lindsayplatt:
Just FYI, this was merged into a branch. Do we want to keep it there for now? Not sure what that does to the pipeline.

aappling-usgs:
Is there any good reason to keep it in a branch? If all it does is add new files to the pipeline, and if the master branch never tries to create files of that name, then the master branch pipeline won't suffer...but the master branch also won't have these potentially useful files, and if there's file overlap then there could be unnecessary rebuilds. I'm also generally against having parallel branches without strong reason because they require human monitoring and I don't usually have the brain space to keep track of them, personally.

lindsayplatt:
It doesn't duplicate any of the filenames, just adds new ones. Not sure what you mean by `the master branch also won't have these potentially useful files` - are you talking about if we do not merge this branch into master?

I am also generally for keeping one branch, unless parallel branches are a necessity to the workflow, such as with `vizstorm-gif`.

aappling-usgs:
> are you talking about if we do not merge this branch into master
yes, that's what I mean

I agree about vizstorm - that's the one case I've seen so far where persistent branches have made sense to me

jread-usgs:
I had it as a branch because I was writing some junk code in advance of the meeting and wanted it to be clear that it wasn't for prime time...
I'd say we try to get this into master IMO

lindsayplatt:
Does #135 need to be done in order for this to be master?

jread-usgs:
No, I think we can get it in now but that is kind of a broken-window issue that we should do soon. Alison and I have the Kw_file(s) so would need to be one of us or I can share w/ you Lindsay if you have bandwidth.

PR #142, opened 2020-03-05
hypsos for MI

### LINKED ISSUE #141: Add MI hypso

jread-usgs:
fixes #141


4_params_munge/src/munge_mglp.R
```diff
@@ -11,3 +11,33 @@ mglp_depths_from_TDOx <- function(out_ind, TDOx_ind, mglp_xwalk_ind){
...
+  mglp_xwalk <- readRDS(sc_retrieve(mglp_xwalk_ind)) %>%
+    left_join(areas, by = 'MGLP_ID')
+
+  # lakes like MIstjom36571 have >1 name, eg  Havens Lake & Goodrich Lake
+  # Two Hearted Lak has multiple long_dd and lat_DD
+
+  bathy_areas <- scipiper::sc_retrieve(bathy_csv_ind) %>% read_csv %>%
+    mutate(MGLP_ID = paste0('MGLP_', MGLP_ID), depths = Depth_ft * 0.3048) %>%
+    inner_join(mglp_xwalk) %>% mutate(areas = LakeAfr_Be * areas_m2) %>%
+    mutate(other_ID = paste0(MGLP_ID, LAKE_NAME, LONG_DD, LAT_DD)) %>%
```
lindsayplatt: Is this the part that collapses some of the MI lakes into one? Seems like this would do the opposite because there are now more unique "other_IDs" than "MGLP_IDs"? Since you noted that some MGLP_ID have multiple LAKE_NAME values.

4_params_munge/src/munge_mglp.R
```diff
@@ -11,3 +11,33 @@ mglp_depths_from_TDOx <- function(out_ind, TDOx_ind, mglp_xwalk_ind){
...
+
+
+munge_mglp_mi_perc_bathy <- function(out_ind, bathy_csv_ind, mglp_xwalk_ind, lake_poly_ind){
+
+
+  lakes <- scipiper::sc_retrieve(lake_poly_ind) %>% readRDS()
+  # multiple matches to
+
+  areas <- data.frame(MGLP_ID = lakes$site_id, areas = as.numeric(st_area(lakes)), stringsAsFactors = FALSE) %>%
+    group_by(MGLP_ID) %>% summarize(areas_m2 = sum(areas))
```
lindsayplatt: Can you describe why you would need to do this step - grouping and then summing areas? Are there multiple lake polygons for individual MGLP_IDs?

4_params_munge/src/munge_mglp.R
```diff
@@ -11,3 +11,33 @@ mglp_depths_from_TDOx <- function(out_ind, TDOx_ind, mglp_xwalk_ind){
...
+
+  mglp_xwalk <- readRDS(sc_retrieve(mglp_xwalk_ind)) %>%
+    left_join(areas, by = 'MGLP_ID')
+
+  # lakes like MIstjom36571 have >1 name, eg  Havens Lake & Goodrich Lake
+  # Two Hearted Lak has multiple long_dd and lat_DD
+
+  bathy_areas <- scipiper::sc_retrieve(bathy_csv_ind) %>% read_csv %>%
+    mutate(MGLP_ID = paste0('MGLP_', MGLP_ID), depths = Depth_ft * 0.3048) %>%
+    inner_join(mglp_xwalk) %>% mutate(areas = LakeAfr_Be * areas_m2) %>%
```
lindsayplatt: I do not understand what `areas = LakeAfr_Be * areas_m2` does. Can you add a comment here?

4_params_munge/src/munge_mglp.R
jread-usgs: `other_ID` is the field that has 1 or more values per unique `site_ID`. In most cases, (e.g., MNDOW) this is a state identifier. In this case, I had to create my own ID that was a mashup of the different properties of the lake since there was no actual unique ID. We should be able to take a closer look in #143

4_params_munge/src/munge_mglp.R
jread-usgs: This isn't actually needed...so I removed it in my recent commit.

PR #147, opened 2020-03-26
Update summary figure to acknowledge the digitizing work being done

lindsayplatt:
Holly and I wanted an easy way to let the undergraduates know which lakes to focus on finding hypso data. She sent a list of lakes that we have known hypso for and are being digitized. I included this information in the summary figure now.

Lots of data, but we still need to hunt down hypsography:
image

Lots of data, but we have hypsos being digitized:
image

Lots of data, but no hypsos in the pipeline:
image


jread-usgs:
while you are in there, can you add the (n={number}) text to the legend to mimic what exists in the TOHA summary html? That info is helpful. Looks good, and my suggestion is optional if it is a pain to implement.

lindsayplatt:
Reordered to what I thought was a more logical progression. Let me know if you'd like to see a different order:

image

jread-usgs:
These numbers don't look quite right. 0 GLM lakes? That number should be 6.7K or so...

lindsayplatt:
although ... just realized this is a bit sketchy with PGDL, TOHA, and GLM ...

lindsayplatt:
It is due to the filter deciding there is not any meteo data. I think someone else will have to build after this is merged (or build and PR to my open PR) because it is looking in `7_drivers_munge/out` for files that I don't have and only someone who has built that step would have.

https://github.com/USGS-R/lake-temperature-model-prep/blob/master/8_viz/src/merge_lake_data.R#L16-L17

jread-usgs:
ahhhhh, gotcha. Ok, I need to build

lindsayplatt:
hold on a sec - made a tiny change that is coming

lindsayplatt:
Ok, done

jread-usgs:
Merged and rebuilt, looks good:

image

PR coming with the ind and build updates

lindsayplatt:
@jread-usgs are you planning to build before or after this is merged?

Issue #100, opened 2019-11-07
update cell NLDAS files to include up to date canonical lakes


jread-usgs:
See issue in #99 as to why this is important. Currently the lakes modeled where only lakes that existed in the spatial files before the switch to NHD HR. This was a point in the workflow where we had added lakes beyond the Winslow release, so we do have additional states covered. We're still missing some easy adds for new lakes that we picked up with the switch to NHD HR. 

But, we're not in a position where it would make sense to rebuild the .nc cube part of this workflow, which is very time consuming and seems to have troubles (jeff and I couldn't run it on mac/nix/win after I upgraded my OS from #1 instructions). However, the "cell" part of this workflow should be fine and fast if we want to simply run that to pick up the ~500 or so missing lakes (more if we don't include secchi as a minimum data requirement).

jread-usgs:
I am doing this currently for Lake of the Woods. It will result in building more NLDAS cells for the TOHA release, and then generate a new driver file for LotW. This required very minor changes to 6_drivers_fetch.yml

PR #149, opened 2020-03-29
update w/ Lake of the Woods drivers

jread-usgs:
NLDAS cells for the MN "notch" were built in https://github.com/jread-usgs/lake-temperature-model-prep/tree/lake_o_woods, and because there is no direct coupling to the driver files (we simply check what exists in a local directory), I am updating here w/o PR'ing that other branch. That other branch changes the NLDAS footprint to a small part of northern MN, so I don't want that in the canonical repo/branch. But this viz update is relevant since it tells us what files exist and don't exist for modeling. We've now added one high profile lake: Lake of the Woods see #100


jread-usgs:
These are just build and ind updates, so merging my own.

PR #174, opened 2020-05-24
Update for "LAB" data filtering

jread-usgs:


jread-usgs:
nothing wild in here, so merging my own. We've got the `"LAB"` site filtering in here and then a full rebuilt, so some status files and hashes.

PR #178, opened 2020-10-08
Completely clean build on this pipeline; adding DRB

jread-usgs:
@aappling-usgs DRB NLDAS drivers are now in the pipeline

@limnoliver can you take a look at [these deleted files](https://github.com/USGS-R/lake-temperature-model-prep/commit/c11aa674b92c23b252c302d08cdab8de97490fe0) and let me know if any raise eyebrows

Overall I don't think there are many real changes in here, other than lots of cleanup and lots of rebuilds. I did a completely fresh build of the whole pipeline.


limnoliver:
I don't think the [build](https://github.com/USGS-R/lake-temperature-model-prep/commit/c11aa674b92c23b252c302d08cdab8de97490fe0#diff-00b7d8e43559b8a1511842148a103cb8) file for `7b_temp_merge/out/source_metadata_for_release.csv.ind` should be deleted. Though it has a slightly older build date so not sure why that would not have been rebuilt?

jread-usgs:
Yeah, I bet that one wasn’t rebuilt because it isn’t a dependency for 8_viz? Since I wiped all builds and inds, it would appear as deleted if that target didn’t get rebuilt.

limnoliver:
The LTER build files also did not get rebuilt, which I _think_ makes sense, but we may want to rethink down the road. Those are in the coop yml, but serve to download from the NTL websites and then push to the coop location in drive so they are included in the parsing step. Perhaps these should be added in `depends:` for [coop_all_files.rds.ind](https://github.com/jread-usgs/lake-temperature-model-prep/blob/DRB_reservoirs/6_temp_coop_fetch.yml#L49)?

limnoliver:
These two are confusing to me - [6_temp_coop_fetch/out/coop_all_files.rds.ind](https://github.com/USGS-R/lake-temperature-model-prep/commit/c11aa674b92c23b252c302d08cdab8de97490fe0#diff-9154ebf2bbaedb8c55173a9998b88249) and [`6_temp_coop_fetch/log/6_temp_coop_fetch_tasks.ind`](https://github.com/USGS-R/lake-temperature-model-prep/commit/c11aa674b92c23b252c302d08cdab8de97490fe0#diff-f4dc9f216de396d31d7b2c797789815d) did not get rebuilt, which seems impossible since you rebuilt the coop files?

jread-usgs:
I don't understand the coop .ind files either. hmmm. As for the LTER files, I looked into this and I understand why they didn't build but I think we may want to insert them into the pipeline in a different way in the future.

jread-usgs:
For the first, I was able to create a build file by forcing a build. For the second, I tried the same and got
```
Error in target_run(target, store, quiet) : 
  command for 6_temp_coop_fetch/log/6_temp_coop_fetch_tasks.ind did not create file
```
for 
```r
scmake('6_temp_coop_fetch/log/6_temp_coop_fetch_tasks.ind', force = TRUE)
```
Not sure what's going on there and I do not have a chance to diagnose right now.

limnoliver:
This is vaguely making me recall a scipiper gotchya that maybe @aappling-usgs remembers. I did find this text below, and recall something about special things involved when there's a .promise. Did you use `scdel()` to remove those files?

* To force a rebuild, either use the `force=TRUE` argument to `scmake()` or use `scdel()` to delete <ind>**indicator files**</ind>. There's seldom any benefit to deleting <data>**data files**</data> (by any method); usually deleting the <ind>**indicator files**</ind> is plenty. When deleting <ind>**indicator files**</ind>, `force=TRUE` or `scdel()` are preferable to directly deleting the `.ind` files because if only the `.ind` files are deleted, the `scipiper` database may fail to update properly when the `.ind` files are rebuilt.

jread-usgs:
I bet you are right. That .ind is the result of a task table (I think) so the promise needs to be `scdel`'d from inside the task yaml. All of this is cleared up with the newer method of teaching task tables per the training, but this pipeline uses the old method that is subject to this gotcha. Haven't verified though. Will during nap time ;)

jread-usgs:
This is complete now. I am merging.

aappling-usgs:
Nice diagnosis and solution, you two! Sounds just right.

PR #179, opened 2020-10-09
DRB reservoir depth, temperature, and secchi adding; GLM model read

jread-usgs:
This is on top of #178 so shouldn't be merged before that one. 

My main goal was to see if we could get GLM-ready data out of the pipeline for a new state:

image

- I added NY to the list of states to pull in the NHDHR pull
- I modified the bounding box for the WQP pull (then had to hack the code inside to split the box in two because it was timing out...)
- I added a simple target in 4_config that allowed me to specify a max depth for a given `site_id`, and used that to add a few reservoirs (e.g., `site_id` came from `readRDS('2_crosswalk_munge/out/gnisname_nhdhr_xwalk.rds') %>% filter(GNIS_Name == 'Neversink Reservoir')`)
- ran `scmake('8_viz')`


jread-usgs:
@aappling-usgs can you take a look? I know I've got the hacky bbox thing in there 🤮.

1_crosswalk_fetch/src/fetch_crosswalk.R
```diff
@@ -214,12 +214,27 @@ fetch_wqp_lake_sites <- function(ind_file, characteristicName, bBox, dummy){
...
           hacking resultCount as 200')
   # when switching back in the future, seems lat and lon were the names for LatitudeMeasure & LongitudeMeasure
-  lake_sites_sf <- whatWQPsites(siteType = "Lake, Reservoir, Impoundment", characteristicName = characteristicName,
-                                bBox = bBox) %>%
+
+  # this is breaking with a single bbox query when adding DRB...uggg
+  bbox1 <- bBox
+  bbox1[3] <- bbox1[1] + diff(bBox[c(1,3)])/2
+  bbox2 <- bBox
+  bbox2[1] <- bbox1[3]
```
aappling-usgs, line 222: This seems like a reasonable start on a general fix for too-big bboxes: sure, split it in half! And I think we can make it more flexible in the future as we add more locations.

1_crosswalk_fetch/src/fetch_crosswalk.R
```diff
@@ -214,12 +214,27 @@ fetch_wqp_lake_sites <- function(ind_file, characteristicName, bBox, dummy){
...
     dplyr::select(site_id = MonitoringLocationIdentifier, resultCount, LatitudeMeasure, LongitudeMeasure) %>%
     st_as_sf(coords = c("LongitudeMeasure", "LatitudeMeasure"), crs = 4326)
 
+  lake_sites_sf2 <- whatWQPsites(siteType = "Lake, Reservoir, Impoundment", characteristicName = characteristicName,
+                                 bBox = bbox2) %>%
+    mutate(resultCount = 200) %>%
+    dplyr::select(site_id = MonitoringLocationIdentifier, resultCount, LatitudeMeasure, LongitudeMeasure) %>%
+    st_as_sf(coords = c("LongitudeMeasure", "LatitudeMeasure"), crs = 4326)
+
+  lake_sites_sf <- rbind(lake_sites_sf1, lake_sites_sf2)
```
aappling-usgs, line 236: Unless `whatWQPsites` returns inconsistently sized/formatted data.frames, I'd think you could just call lines 224-225 and 230-231, then `rbind` those, then run lines 226-228 just once rather than duplicating those lines in 232-234.

1_crosswalk_fetch/src/fetch_crosswalk.R
aappling-usgs: How long do these WQP calls take? If it's a long time, you might consider a custom split to make a small bbox around the DRB so we're not pulling all the data from the Northeast. We'll have to revise this code again for more generality if and when the domain grows again, anyway, so I don't see any harm in making it fairly specific to today's challenge for now.

aappling-usgs: But also, if it's already pulled, let sleeping dogs lie.

4_params_munge.yml
```diff
@@ -37,12 +37,19 @@ targets:
...
 
   4_params_munge/out/toha_varying_kw.rds.ind:
     command: munge_daily_secchi(target_name,
       kw_files_zip_ind = '3_params_fetch/in/USGS_Kd_Predictions_byNHDHR.zip.ind')
 
+  # nhdhr_151957878: Pepacton Reservoir; nhdhr_120022743: Cannonsville Reservoir; nhdhr_149269381: Neversink Reservoir; nhdhr_61790779: Rondout Reservoir
+  4_params_munge/out/NY_manual_depths.rds.ind:
+    command: munge_simple_depth(target_name, nhdhr_151957878 = 55, nhdhr_120022743 = 42.672, nhdhr_149269381 = 53, nhdhr_61790779 = 53.3)
+  4_params_munge/out/NY_manual_depths.rds:
+    command: gd_get('4_params_munge/out/NY_manual_depths.rds.ind')
```
aappling-usgs, line 51: In other repositories we've switched over to a system of putting all `gd_get`/`s3_get` calls in a separate `getters.yml`, which substantially cuts down on unnecessary rebuilds. we should probably do this in lake-temperature-model-prep too.

4_params_munge/src/munge_bathy.R
```diff
@@ -16,6 +16,18 @@ munge_wbic_bathy <- function(out_ind, bathy_zip_ind, wbic_xwalk_ind){
...
+
+  data_file <- scipiper::as_data_file(out_ind)
+
+  data.frame(...) %>%
+    pivot_longer(cols = starts_with('nhdhr_'),
+                 names_to = "site_id",
+                 values_to = 'z_max') %>%
+    saveRDS(file = data_file)
+  gd_put(out_ind, data_file)
+}
```
aappling-usgs, line 29: Nice, this makes for a really clean command in the YAML

6_temp_coop_fetch.yml
```diff
@@ -18,8 +18,8 @@ targets:
   6_temp_coop_fetch:
     depends:
       - 6_temp_coop_fetch/downloads/long_term_ntl.csv.ind
-      - 6_temp_coop_fetch/downloads/mendota_daily_buoy.csv.ind
-      - 6_temp_coop_fetch/downloads/mendota_temps_long.csv.ind
+      # THIS IS FAILING TO DOWNLOAD - 6_temp_coop_fetch/downloads/mendota_daily_buoy.csv.ind
+      # THIS IS FAILING TO DOWNLOAD - - 6_temp_coop_fetch/downloads/mendota_temps_long.csv.ind
```
aappling-usgs, line 22: hmm, I'm able to browse to https://lter.limnology.wisc.edu/sites/default/files/data/ntl130_1_v5.csv, https://lter.limnology.wisc.edu/sites/default/files/data/ntl130_1_v7.csv, and https://lter.limnology.wisc.edu/sites/default/files/data/Lake_Mendota_temps_long.csv. At the moment. Is it possible the site was just down temporarily?

aappling-usgs (approved):
Looks good! I added a few discussion points and possible changes for next time, but I don't see any real need for revisions this time through.

1_crosswalk_fetch.yml
```diff
@@ -100,13 +100,13 @@ targets:
 
   1_crosswalk_fetch/out/wqp_lake_temperature_sites_sf.rds.ind:
     command: fetch_wqp_lake_sites(target_name, characteristicName = wqp_temperature_charnames,
-      bBox = I(c(-105.79725, 34.95527, -78.85268, 50.12992)), dummy = I('2020-01-13'))
+      bBox = I(c(-105.79725, 34.95527, -74, 50.12992)), dummy = I('2020-01-13'))
```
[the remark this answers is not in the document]
jread-usgs: Good point. I don't like this function right now as it is a hack and the prior one did indeed need the dummy. When WQP stopped being able to handle a sites query w/o a spatial filter (like bbox), this function got more brittle, because the arguments now need to be updated when you modify your spatial area of interest (e.g., add NY). I will update the dummy and make the other change, but would like to come up with a generic way of getting all US sites in a mosaic pull. To answer your question below, no, the site query doesn't take too long.

1_crosswalk_fetch/src/fetch_crosswalk.R
jread-usgs: yes. I was sleepy/lazy

6_temp_coop_fetch.yml
jread-usgs: It is odd. I noticed this in the LAGOS fetches too - the pasta (the underlying system that EDI/LTER uses) doesn't seem to work with `download.file` or `GET` for these files, even though the browser downloads work. I didn't mess around with trying to figure out why and I also think these files should be in `in` instead of a web fetch here.

PR #185, opened 2020-11-16
Wqp tasks

jread-usgs:
Updating task remaking pattern per discussion in #pipelines


1_crosswalk_fetch/src/fetch_wqp_sites.R
```diff
@@ -0,0 +1,86 @@
+fetch_wqp_sites <- function(out_ind, characteristicName, sf_ind, box_res, dummy, ...) {
+
+  target_name <- scipiper::as_data_file(out_ind)
+  remakefile <- basename(target_name) %>% tools::file_path_sans_ext() %>% paste0('_tasks.yml')
+  sources <- c(...)
+
+  sf_geometry <- readRDS(scipiper::sc_retrieve(sf_ind))
+
+  bbox_grid <- sf::st_make_grid(sf_geometry, square = TRUE, cellsize = box_res, offset = c(-180,15))
```
limnoliver, line 9: Cool, did not know this function existed.

6_wqp_fetch/src/fetch_wqp_data.R
```diff
@@ -0,0 +1,145 @@
...
+  message(sprintf("**dropping %s sites and %s results due to containing 'RCE WRP-'", nrow(bad_RCE), sum(bad_RCE$resultCount)))
+  message(sprintf("**dropping %s sites and %s results due to containing 'ALABAMACOUSHATTATRIBE.TX_WQX'", nrow(bad_ALABAMACOUSHATTA), sum(bad_ALABAMACOUSHATTA$resultCount)))
+
+  # now filter out bad sites
+  wqp_site_table <- filter(wqp_sites, !MonitoringLocationIdentifier %in% c(
+    bad_char$MonitoringLocationIdentifier,
+    bad_RCE$MonitoringLocationIdentifier,
+    bad_ALABAMACOUSHATTA$MonitoringLocationIdentifier)) %>%
+    arrange(desc(resultCount)) %>%
+    mutate(task_num = MESS::cumsumbinning(x = resultCount, threshold = max_results, maxgroupsize = max_sites),
```
limnoliver, line 24: oooh, nice find!

6_wqp_fetch/src/fetch_wqp_data.R
```diff
@@ -0,0 +1,145 @@
...
+           task_name = paste0("pull_", task_num))
+
+
+  wqp_site_file <- paste0(pull_type, '_site_table.rds') %>% file.path(data_tmp_dir, .)
+  saveRDS(wqp_site_table, file = wqp_site_file)
```
limnoliver, line 29: This file isn't tracked by scipiper?

limnoliver (commented):
Looks great, Jordan! I think the bbox is a nice and seemingly straightforward approach.

6_wqp_fetch/src/fetch_wqp_data.R
jread-usgs: No, this one isn't. I wrote to a file in this temp directory so that I could fully decouple the task makefile from the parent makefile (no `includes`) which makes it a bit simpler to code, but introduces this artifact that sticks around. 

I think I will just delete this file when the function exits, since it is written fresh every time the function is called, regardless of whether its dependency changed.

jread-usgs:
Merging. Dealt with the review comment in a way that I think makes sense. An alternative would be to link the task remake file with the parent w/ `include`, but I was avoiding that for simplicity.

Issue #184, opened 2020-11-15
Find a way to re-claim sites dropped because their names contain `/`


jread-usgs:
We do WQP pulls based on a `POST` body that includes `siteid` as the `MonitoringLocationIdentifier`. Many of these `MonitoringLocationIdentifier` have "/" in the name, which causes a query error and because of that we have been pre-filtering out these site from the queries. But we may be losing key sites because of this, for example a recent Secchi data pull has this: `**dropping 301 sites and 13885 results due to containing '/'`

It is likely that many of these sites are under the same local `organization`, and as such, could be queried based on that ID and then the results could be filtered back to contain only the siteids that were skipped from the normal query. 

Before we embark on this type of fix, we'd want to have a better understanding of what we'd gain. Are we missing key, well-studied lakes? Or are the sites we are excluding not that big of a deal?

jread-usgs:
One example of this is for secchi data (`c("Depth, Secchi disk depth", "Depth, Secchi disk depth (choice list)", "Secchi Reading Condition (choice list)", "Water transparency, Secchi disc")`), where the "UMC" `OrganizationIdentifier` has 13,562 results and 279 unique sites that get tossed because of bad site identifiers (such as this "UMC-7067/1.7"). 

Seems we can get these data by specifying `organization` instead of `siteid` and then filtering sites after the data are returned.

PR #186, opened 2020-11-18
rescue on some of the bad monitoring IDs

jread-usgs:
See #184. 

This creates additional tasks for organizations that have bad monitoring IDs and adds those pulls to the task table.


jread-usgs:
@limnoliver small extension of what you already reviewed

6_wqp_fetch/src/fetch_wqp_data.R
```diff
@@ -136,21 +134,24 @@ pull_wqp_data <- function(data_file, wqp_sites, characteristicName, dummy){
...
+  } else {
+    wqp_args$siteid <- pull(wqp_sites, MonitoringLocationIdentifier)
   }
 
 
   # do the data pull
   # first pull using readWQPdata, then if that fails, try POST
   wqp_dat <- suppressMessages(wqp_POST(wqp_args))
 
+  if (wqp_sites$pull_org){
```
limnoliver, line 152: Do you need a subset here, as above (`wqp_sites$pull_org[1]`)?

6_wqp_fetch/src/fetch_wqp_data.R
```diff
@@ -136,21 +134,24 @@ pull_wqp_data <- function(data_file, wqp_sites, characteristicName, dummy){
...
+    wqp_dat <- filter(wqp_dat, MonitoringLocationIdentifier %in% c(wqp_sites$MonitoringLocationIdentifier))
```
limnoliver, line 153: small note - the c() is not needed here?

limnoliver (approved):
Looks good, Jordan -- glad you were able to save those sites!

6_wqp_fetch/src/fetch_wqp_data.R
jread-usgs: correct. Will fix

6_wqp_fetch/src/fetch_wqp_data.R
jread-usgs: yes, this one needs that (same as above), but works like this just with an warning that it is using the first of the vector. Will update this one too.

Issue #189, opened 2020-12-09
brittle link to expanding states


jread-usgs:
This caught us in #81 

To expand the NHD states in the pull, we currently use

```yaml
  nhd_HR_download_plan:
    command: create_nhd_HR_download_plan(I(c("North_Dakota","South_Dakota","Iowa","Michigan","Indiana","Illinois","Wisconsin",
      "Minnesota", "New_York", "Pennsylvania", "Missouri")),
      min_size = I(40000))
```
(but see #188 ) to specify states. But in other targets:

```yaml
  1_crosswalk_fetch/out/LAGOS_lakes_sf.rds.ind:
    command: LAGOS_zip_to_sf(
      out_ind = target_name,
      layer = I("LAGOS_NE_All_Lakes_4ha"),
      zip_ind = '1_crosswalk_fetch/in/LAGOS_NE_All_Lakes_4ha.zip.ind',
      states = I(c("IA", "IN", "SD", "ND", "MI", "MN", "WI", "IL")))

  1_crosswalk_fetch/out/MGLP_lakes_sf.rds.ind:
    command: MGLP_zip_to_sf(
      out_ind = target_name,
      gdb_file = I("MGLP_model_data&results.gdb"),
      zip_ind = '1_crosswalk_fetch/in/MGLP_lakes.zip.ind',
      states = I(c("IA", "IN", "SD", "ND", "MI", "MN", "WI", "IL")))
```

This is too many places to specify states, and it is brittle. We should link these things appropriately.

Issue #163, opened 2020-04-08
Add other sources of Kw and secchi


jread-usgs:
Right now the `munge_Kw` function only takes secchi data from the WQP. Seems we'll have plenty of other sources. Make this more generic in the future to accept all kinds in ranked priority (see how the `H_A_values` are calculated w/ inputs for example)

jread-usgs:
comment from #111 (now closed, since it is redundant) 
> Now that we have WBIC xwalk, this should be fairly easy? https://github.com/USGS-R/necsc-lake-modeling/tree/d37377ea422b9be324e8bd203fc6eecc36966401/data/secchi_data_raw

jread-usgs:
Seems we're not even including LAGOS secchi here, which is in the pipeline but not linked to the production of the `kw` target

jread-usgs:
Now we'll want to add the landsat secchi source from the Rose lab. Likely `1_crosswalk_fetch/in`

jread-usgs:
Added `Max_Glines_LANDSAT_secchi_medians.csv` to the 1_crosswalk_fetch/in in case we want to get going on this @lindsayplatt

jread-usgs:
Connected to #193

lindsayplatt:
Another source (come Fall 2022) could be expanding our LAGOS source of water clarity from LAGOS-NE to LAGOS-US via the release of LAGOS-US-LIMNO (see [here](https://lagoslakes.org/lagos-us-overview/) for details). This was originally proposed via #220.

PR #191, opened 2020-12-10
Adding LAGOS secchi and LAGOS depth data

jread-usgs:
Part of #163 , but still clean up needed. 

didn't work out #189, but this addresses the oversight that that issue is related to. Still brittle though.


jread-usgs:
Merging this and will create an issue to capture the brittle code part for clean up later. Since this PR impacts gdrive and Lindsay is working in this repo, I want it to be current.

PR #198, opened 2021-02-09
Adding full 2020 times too (DRB only!)

jread-usgs:


jread-usgs:
Pretty straightforward on this one. Time domain change only.

aappling-usgs (approved):
Hi Jordan, I've looked through this PR and made the request for updated `meteo_fl` in the nml_list (which you met in ec8b089). My understanding of this repo is shallow, and there are a lot of changed files, so I may very well have missed something, but I don't see anything else of concern. I was able to use the two new meteo files for DRB Cannonsville and Pepacton.

I did have to update my downstream code to expect 15320 rather than 14975 rows in each meteo file. You may have to make a similar change in other downstream repos. There were no duplicated dates in the particular files for Cannonsville or Pepacton.

You've asked me not to merge because this change only affects DRB, and therefore many other lakes are out of date with respect to the nml file ("would like to keep the PR open until I update the other cubes to this time domain"). So I'll just mark this Approved and move on.

jread-usgs:
Thanks Alison - I am comfortable with the changed files. I am adding the time update to the other "cubes" in this modeling domain and will add another commit that includes the full time update (beyond just the DRB), then will merge.

jread-usgs:
merging as these cover the other parts of the footprint now too. 

Lots of ind/build file changes, but only a few tiny code changes related to NLDAS URL and the time period for NLDAS pulls.

PR #201, opened 2021-04-26
Adding NYC DEP Reservoir Temperature Data 

RAtshan:
Processing and adding Cannonsville and Pepacton reservoirs data that is provided from NYC-DEP into the pipeline. 

First, processing the NYC_DEP reservoir temperature data to retrieve a single temperature profile per reservoir per date. Then, combining the processed NYC_DEP_reservoir_temperature data with the processed  drb_reservoir_temp data. Finally, processing the NYC_DEP reservoir temperature data to extract the water surface level elevation data to retrieve a single surface elevation per reservoir per date.


7a_nwis_munge/src/munge_nyc_dep_reservoir_data.R
```diff
@@ -0,0 +1,61 @@
...
+    filter(n_dates == max(n_dates)) %>%
+    ungroup()
+
+  # Selecting the site with maximum n_dates.
+  sites_max_dates <- unique(max_dates$source_id)
+
+  # Filtering the data to one profile per day per reservoir per site.
+  daily_profile <- dat %>%
+    # Filtering to the sites with the maximum n_dates
+    filter(source_id %in% sites_max_dates) %>%
```
limnoliver, line 44: this is filtering to all sites that had the maximum number of observations on any given data, which will keep _all_ of that site's observations, not just those dates.

7b_temp_merge/src/merge_all_temp_dat.R
```diff
@@ -276,27 +276,35 @@ reduce_temp_data <- function(outind, inind) {
...
            time_hm = lubridate::hm(time),
            hours_minus_noon = time_hm$hour - 12,
            minutes_decimal = time_hm$minute/60,
            hours_diff_noon = abs(hours_minus_noon + minutes_decimal)) %>% arrange(site_id, date, depth_category)
 
-  dat_out <- dat %>%
+  nyc_dep_reservoirs_temps <- readRDS(sc_retrieve(nyc_dep_ind))
+
+  # Selecting unique dates in nyc_dep_reservoirs_temp data.
+  nyc_dep_dates <- unique(nyc_dep_reservoirs_temps$date)
```
limnoliver, line 293: You have the same issue here as we did before, where you're grabbing dates from both reservoirs, when in reality we want to exclude the dates by reservoir.

7b_temp_merge/src/merge_all_temp_dat.R
```diff
@@ -276,27 +276,35 @@ reduce_temp_data <- function(outind, inind) {
...
     group_by(site_id, date, depth_category) %>%
     slice_min(hours_diff_noon) %>%
     ungroup() %>%
     dplyr::select(site_id, source_id, date, depth, temp)
 
-  saveRDS(dat_out, as_data_file(outind))
-  gd_put(outind)
+  drb_source <- unique(daily_drb_dat$source_id)
 
+  combine_dat <- bind_rows(daily_drb_dat, nyc_dep_reservoirs_temps) %>%
```
limnoliver, line 304: take a look at the `.id` argument for bind_rows, which should help you create a column that tracks the source of the data (so handles what you're trying to do in line 305)

limnoliver: It's okay to leave this as-is, the above comment was to show an alternative method.

.gitignore
```diff
@@ -71,28 +59,21 @@ rsconnect/
...
 # ignore bathy downloads
 3_params_fetch/in/hypsos_m/*
-
 # ignore html files
 *.html
-
 # ignore source metdata csv
 7b_temp_merge/in/source_metadata.csv
 7b_temp_merge/out/source_metadata_for_release.csv
+sentinel_lake_names.csv
```
limnoliver, line 79: Changes in this doc are not relevant to this PR. Looks like some line ending issues + an unrelated change?

RAtshan: It is unrelated change. I added the file that was setting in the my local project to the .gitignore file.

limnoliver:
Rasha - can you do some quality checks on these data? These temperature values should be in celsius, and generally should range from 0-32. I'm seeing some unrealistic values when I read in the final dataset:

```
dat <- readRDS(sc_retrieve('7b_temp_merge/out/drb_daily_reservoir_temps.rds.ind'))
summary(dat$temp)
#Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
#-110.00    5.81   10.13   11.77   17.21   93.28 
```

Issue #203, opened 2021-05-10
Fix multiple obs per reservoir-date-depth


limnoliver:
Jordan noted there are multiple observations per reservoir-date-depth in the [Cannonsville/Pepacton temperature dataset](https://github.com/USGS-R/lake-temperature-model-prep/blob/master/7b_temp_merge.yml#L76). Investigate and get down to a single obs per reservoir-date-depth.

RAtshan:
After investigating the NYC_DEP reservoir raw data, I found that there're multiple temperature measurements per reservoir-date-depth. To fix the issue I found the average temperature per reservoir-source-date-depth. (#204 )

PR #204, opened 2021-05-12
Fix multiple observation per reservoir-date-depth Issue

RAtshan:
Fixing multiple observation per reservoir-date-depth issue (#203 ). I noticed that there're different temperature measurements for the same reservoir, date, and depth that were taking during different time stamps. To fix the issue I found the average mean of the temperature that are taken from a reservoir on date at particular depth.


limnoliver:
I merged from my version of Rasha's Branch, so closing this one.

PR #208, opened 2021-06-24
full build after full driver update

jread-usgs:


jread-usgs:
All just build and .ind updates, so merging my own.

Issue #45, opened 2019-04-30
cooperator data munge in support of expansion to MGLP footprint


limnoliver:
A list of munging tasks that need to be done to incorporate cooperator data in the expanded footprint (part of #33). 
 
SOUTH DAKOTA

- [ ] Crosswalk between sites in file `SD_Lake_temp_export.xlsx` and our IDs (in this case, MGLP IDs). There is lat/long and good metadata for each lake to do some quality checking. 

- [ ] Track down depth data for `SD_Lake_temp_export.xlsx`. There are zero values for maximum depth for some lakes in tab `Metadata_Max_Depth`. I found a few depth resources for SD: a [thesis ](https://openprairie.sdstate.edu/cgi/viewcontent.cgi?article=1218&context=etd) and a [USGS report](https://pubs.usgs.gov/wri/1984/4030/report.pdf). Or might be able to get some site metadata from the [org's data portal](https://apps.sd.gov/NR92WQMAP/). 

INDIANA

- [ ] Crosswalk between sites and our IDs (in this case, MGLP IDs). This is likely best done using lake metadata in file `Indiana CLP lake data 1994-2013`. In this file, there is lat/long and county info to help ID water bodies. This should cover many lakes in file `Indiana_Glacial_Lakes_WQ_IN_DNR` and `Indiana_GlacialLakes_TempDOprofiles_5.6.13`. 

- [ ] Verify data lakes have depth info. Depth data are included in `Indiana CLP lake data 1994-2013`. 

OTHER
- [ ] Additional cooperator files are currently not being incorporated. These include anything in a sub-folder here. Note that many of these lakes already have data (e.g., the PCA subfolder and Ten Mile Lake subfolder) but in some cases, I know these will add to the record (e.g., Ten Mile missing data in the early 2000s that will be covered by these files).

- [ ] File `Water_Temp.accdb` has negative depths that have not been resolved. The negative depths, according to the explainer file, are ```Water temperature sensor depth. This number is positive if measurement is taken from below the water surface and negative if measurement is taken from lake bottom.``` I confirmed that we cannot easily find what depth from surface this is. That is, -41m is 41 meters from the bottom, and there are no metadata to support figuring out what the site depth is. 

[...]

PR #211, opened 2021-07-02
Starting to add Iowa data. Keeping this open because it impacts the cache and others may need to pull

jread-usgs:


jread-usgs:
Handles part of #45

jread-usgs:
Finally gives us some observations in IA lakes: 
image

7a_temp_coop_munge/src/crosswalk_fxns.R
```diff
@@ -28,6 +29,8 @@ crosswalk_coop_dat <- function(outind = target_name, inind,
   dow2nhd$DOW <- gsub('mndow_', '', as.character(dow2nhd$MNDOW_ID))
   dow2nhd <- distinct(dow2nhd) # get rid of duplicated obs
 
+  iowa2hnd <- sc_retrieve(iowa_crosswalk) %>% readRDS(dowfile) %>% distinct()
```
hcorson-dosch, line 32: Did you mean for this variable name to be `iowa2nhd` rather than `iowa2hnd`?
And if so, should the second pipe action be `readRDS(iowa2nhd)` instead of `readRDS(dowfile)`?

7a_temp_coop_munge/src/crosswalk_fxns.R
```diff
@@ -41,8 +44,11 @@ crosswalk_coop_dat <- function(outind = target_name, inind,
   dat_wbic <- filter(dat, !is.na(WBIC)) %>%
     left_join(wbics_in_dat)
 
-  # dow
+  # iowa
+  dat_iowa <- filter(dat, !is.na(Iowa_ID)) %>%
+    left_join(iowa2hnd)
```
hcorson-dosch, line 49: If change variable name, would need to change here, too

hcorson-dosch (approved):
Everything looks good to me other than a couple of possible typos that I noted

7a_temp_coop_munge/src/crosswalk_fxns.R
jread-usgs: Good catch! Fixed in my updated commit

PR #216, opened 2021-07-31
Build w/ AR from NHD HR

jread-usgs:
Adding AR to the NHD lake files. This repulled all lakes, since the NHD download URL changed.


hcorson-dosch (approved):
Whew that's a lot of rebuilds!! Everything looks good to me.

PR #218, opened 2021-08-20
Add Bull Shoals reservoir, MO

lindsayplatt:


lindsayplatt:
I was able to get passed all of the build issues finally!! One weird thing ... I ran `scmake("7a_temp_coop_munge/out/all_coop_dat_linked.feather.ind")` and strangely get this warning message. Is that the normal amount we have been seeing get dropped?

```r
In crosswalk_coop_dat(outind = "7a_temp_coop_munge/out/all_coop_dat_linked.feather.ind",  :
  Dropping 1118621 temperature observations due to missing NHD ids.
```

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -0,0 +1,85 @@
+
+# Parse the ~300 files within the `Bull_Shoals_Lake_DO_and_Temp.zip` file
+
+parse_Bull_Shoals_Lake_DO_and_Temp <- function(inind, outind) {
+
+  infile <- sc_retrieve(inind, remake_file = '6_temp_coop_fetch_tasks.yml')
```
jread-usgs, line 6: Likely a good reason for this, but wondering why the `remake_file` for this isn't `getters.yml` for this like it would be for other retrieve/gd_get targets.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -0,0 +1,85 @@
...
+  outfile <- as_data_file(outind)
+
+  # unzip the file with all of Bull Shoals
+  files_from_zip <- unzip(infile, exdir = tempdir())
```
jread-usgs, line 10: Not a big deal since the tempdir gets wiped when you close the R session, but might want to get in the habit of doing clean up of the files you extract after you make use of them.

Could do this by using `on.exit()` with a delete/unlink of the unzipped directory whenever the function exists (either due to completing or hitting an error) or using unlink at the end of the function.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -0,0 +1,85 @@
...
+      fix_colname_encoding("latin1") %>%
+
+      # Reformat columns and extract info we need from Excel file
+      dplyr::mutate(DateTime = extract_bullshoals_date(fn),
+                    site = extract_bullshoals_site(fn),
+                    # TODO: Don't manually add the nhdhr id
+                    # Only doing now bc we don't have hypso for
+                    # Bull Shoals yet and I found this nhdhr match:
+                    # https://github.com/USGS-R/lake-temperature-model-prep/blob/master/4_params_munge.yml#L60
+                    id = 'nhdhr_12003253',
```
jread-usgs, line 32: this is probably convention in the `coop_parsers`, but we call this `site_id` not `id` elsewhere in the repo. But no need to change this if the convention is used in this section for some reason.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -0,0 +1,85 @@
...
+                    depth = `Depth (ft)`) %>%
+      dplyr::rename(temp = "Temp (°C)") %>%
+
+      # Replace "Surface" depth and convert depths to numeric
+      mutate(depth = as.numeric(ifelse(depth == "Surface", 0, depth))) %>%
+
+      # Change depth from feet to meters
+      mutate(depth = depth / 3.28) %>%
```
jread-usgs, line 40: elsewhere we multiply feet by 0.3048, which is a bit different than dividing by 3.28. 3.28 is the rounded version of this, and should be more like 3.28084. 

I think it is fine to hard-code it here, but in the future we may want some kind of general "utils.R" set of functions for things like unit conversions that can happen throughout the pipeline and may need to be functions(?). This is such a simple one I don't think it is necessary, but since this conversion is different than how it is done elsewhere, it caught my eye.

jread-usgs (changes requested):
minor change requested on the feet to meters conversion

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
lindsayplatt: I am not sure if there is a good reason, other than to say that I followed the pattern present in all of the other data parsers. None of those use `getters.yml`.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
lindsayplatt: I was following the convention listed [here](https://docs.google.com/document/d/1Z1fg8ng9IWHS9oiQTxsDvpGBlbgz2bMZ5rWNxfRiHC8/edit), where `id` is used for the column that can be crosswalked to an NHDID later.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
lindsayplatt: Are you referring to the `unit_map`s I am seeing elsewhere in the repo? Such as [here](https://github.com/USGS-R/lake-temperature-model-prep/blob/5b934a95f83b9bdcc50fa4704a8817f58de8328b/7a_wqp_munge/src/munge_wqp_dat.R#L12-L14). I see those hardcoded in multiple places, so agree that standardizing would be good. I think Julie is having to convert temperature units and so having helper functions to do that may make sense sooner rather than later.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
jread-usgs: Gotcha, let's keep as-is for that convention. Thanks

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
jread-usgs: Yes, that is one of the places where it shows up.

lindsayplatt: I created #235

lindsayplatt:
@jread-usgs were your concerns addressed or should I make any additional changes before merging?

jread-usgs (approved):

jread-usgs:
Good to merge @lindsayplatt

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
padilla410: I will pick this one off when I add the Norfolk dataset (which will require this conversion)

PR #223, opened 2021-11-10
Targets test

hcorson-dosch:
Set up a simple working example of a targets pipeline within a scipiper pipeline, where the names and hashes of the files generated by the targets pipeline are stored in an indicator file.

### DIFF
--- a/7_drivers_munge/src/GCM_driver_utils.R
+++ b/7_drivers_munge/src/GCM_driver_utils.R
@@ -1,0 +1,12 @@
+library(sf)
+library(ggplot2)
+
+map_centroids <- function(centroids_sf, out_file) {
+
+  centroid_plot <- ggplot() +
+    geom_sf(data=centroids_sf, color='dodgerblue', size=0.5)
+
+  ggsave(out_file, centroid_plot, width=10, height=8, dpi=300)
+
+  return(out_file)
+}
--- a/7_drivers_munge/src/GCM_driver_utils.R
+++ b/7_drivers_munge/src/GCM_driver_utils.R
@@ -1,5 +1,3 @@
-library(sf)
-library(ggplot2)
 
 map_centroids <- function(centroids_sf_rds, out_file) {
   centroids_sf <- readRDS(centroids_sf_rds)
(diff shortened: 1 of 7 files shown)


7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -0,0 +1,12 @@
+library(sf)
+library(ggplot2)
```
lindsayplatt, line 2: I don't think we want these library calls here

lindsayplatt (approved):
This is a great start! Excited to keep building off this.

7_drivers_munge/src/GCM_driver_utils.R
hcorson-dosch: oh whoops, thanks - meant to move those

7_drivers_munge/src/GCM_targets_wrapper.R
```diff
@@ -0,0 +1,10 @@
+build_GCM_pipeline <- function(ind_file, ...) {
+  # Build the targets pipeline
+  tar_make()
```
jread-usgs, line 3: Do you want this to make just one target instead of all? Seems if you include other things in _targets.R that aren't related to this part of the pipeline, you may not want those all to be built or checked.

hcorson-dosch: Yes, that's a good point. I was wondering about that on Wednesday. In this simple example, it works fine, but I agree it would be better to tie this more closely to the final products of our targets pipeline that do relate to this piece of the `lake-temp-model-prep` pipeline (e.g., the eventual processed driver files, the names of which we'll return in the object `pipelines_files_out`). I'll make that change now so that we're set up better for expanding this test pipeline.

hcorson-dosch:
@jread-usgs - going to go ahead and merge since it's a minor change, and so that Lindsay and I can think re: next steps.

7_drivers_munge/src/GCM_targets_wrapper.R
lindsayplatt: Hmmm that's interesting. So setting ourselves up to be able to have multiple `scipiper` targets that call `tar_make()`? That would add some nice flexibility. 

I see @hcorson-dosch made this change, but we probably want to rename now that the target out will be specifically tied to our GCM munging pipeline. I suggest this change in a future PR: `pipeline_files_out` >> `gcm_files_out`

Issue #222, opened 2021-10-22
Add targets pipeline for GCM driver data munging


lindsayplatt:
Create targets pipeline to munge GCM driver data from GDP using `geoknife` to a single (or few) NetCDF file(s). It should work with `2_crosswalk_munge/out/centroid_lakes_sf.rds.ind` as the input for the spatial geom & download data for all 6 of the GCMs (map over these using data.frame of URLs & time periods). We will create a wrapper function that builds this `targets` pipeline and creates the single NetCDF file. The function will be called from the big `scipiper` pipeline.

Plan is to pilot this workflow with 5 lakes, then scale to all of MN, then to the whole footprint.

Considerations:
* Chunking the lakes (1000 at a time is what has been used previously) is a good approach, but we may end up with duplicate data downloads if there are more than one lake in a single cell.
* GDP might be too brittle (if so, may need to adopt the NLDAS approach, see this)
*  Notaro GCMs will need snow, so we should follow a previous approach (to be shared by Jordan later) where snow is created from rain when the air temp is below 0 deg C.
* Could later add in NLDAS to this workflow (rather than where it currently in this pipeline), so that the driver data munging could be pulled out as their own process.

lindsayplatt:
Hayley kicked this off with #223

lindsayplatt:
@jread-usgs some questions for you now that @hcorson-dosch and I are really in the guts of implementing the pipeline.

### Is the format of the data being returned correct?

My current working query ([here is the target](https://github.com/USGS-R/lake-temperature-model-prep/blob/gcm_driver_data_munge_pipeline/_targets.R#L43-L53) and [here is the function](https://github.com/USGS-R/lake-temperature-model-prep/blob/gcm_driver_data_munge_pipeline/7_drivers_munge/src/GCM_driver_utils.R#L49-L64)) returns a messy text file that requires munging – appears to be multiple csv-formatted file separated by headers all in one text file. **Does that seem right or is that triggering alarm bells?** I have started an initial function for parsing but want to make sure before I go much further. Just seems really funky.

Here is a snippet of what is loaded when I do `readLines()`:

```r
[1] "# evspsbl"                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
[...]

PR #224, opened 2021-11-12
Basic geoknife query added to the GCM pipeline

lindsayplatt:
Part of #222 

This adds a few targets to take a subset of `centroid_lakes_sf`, convert to query-able polygons, and download GFDL GCM data as a raw text file. Then, it maps the subset of centroids and the polygons used for the query.

I think merging this basic framework will allow us to work in parallel on 1) the best way to make polygons from our centroids, and 2) munging the text files into a useful format. Later, we will need to work on chunking the centroids in a smarter way and how to do more than one of the GCMs at a time.

Here is the image created:

query_map


hcorson-dosch (approved):
👍 All looks good! Thanks for noting the TODOs in the comments, that's helpful

hcorson-dosch:
Good catch!

PR #225, opened 2021-11-16
Use centroids for GCM query

lindsayplatt:
Change GCM query to work with centroids + use nice geoknife helper, `result()` instead of `download()` to avoid a bunch of parsing work. Currently only working with one variable, but have in issue [here](https://github.com/USGS-R/geoknife/issues/399) to help resolve that.

### DIFF
--- a/_targets.R
+++ b/_targets.R
@@ -6,7 +6,8 @@ tar_option_set(packages = c(
   "sf",
   "scipiper",
   "ggplot2",
-  "geoknife"
+  "geoknife",
+  "arrow"
 ))
 
 source('7_drivers_munge/src/GCM_driver_utils.R')
@@ -35,12 +36,13 @@ targets_list <- list(
   # handle chunks of lakes. See this config file example for an example:
   # https://github.com/USGS-R/necsc-lake-modeling/blob/a81a0e7ed6ede66253f765b42568a4d39a5dccc2/configs/ACCESS_config.yml
   tar_target(
-    gcm_data_raw_txt,
+    gcm_data_raw_feather,
     download_gcm_data(
-      out_file = "7_drivers_munge/tmp/7_GCM_GFDL_raw.txt",
+      out_file = "7_drivers_munge/tmp/7_GCM_GFDL_raw.feather",
       query_geom = query_centroids_geoknife,
       query_url = "https://cida.usgs.gov/thredds/dodsC/notaro_GFDL_1980_1999",
-      query_vars = c("evspsbl", "hfss", "mrso"),
+      # Data definitions: https://cida.usgs.gov/thredds/ncss/notaro_GFDL_2040_2059/dataset.html
+      query_vars = c("evspsbl"), # TODO: more than one var, but geoknife isn't working with that
       query_dates = c('1999-01-01', '1999-01-15')
     ),
     format = "file"
--- a/_targets.R
+++ b/_targets.R
@@ -24,16 +24,10 @@ targets_list <- list(
     split_lake_centroids(centroids_sf_rds)
   ),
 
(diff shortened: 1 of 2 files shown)


lindsayplatt:
@hcorson-dosch here is the start of using centroids to query. This should help you do the next part, which is smartly splitting/grouping and using a constructed grid to query grid centroids, not lakes.

jread-usgs:
Quick follow-up on the geoknife parsing issue. It seems the issue resides with just one of the variables you are using, which is soil moisture (`mrso`). That variable in funky because it has multiple "levels" which are causing some of the assumptions in the parser to fail as captured by Dave [here](https://github.com/USGS-R/geoknife/issues/399#issuecomment-969780589).

```r
gcm_job <- geoknife(
    stencil = simplegeom(query_pts),
    fabric = webdata(
        url = "https://cida.usgs.gov/thredds/dodsC/notaro_GFDL_1980_1999",
        variables = c("evspsbl", "hfss"),
        times = c('1999-01-01', '1999-01-15')
    ),
    wait = TRUE
)
result(gcm_job)
               DateTime         1         2             3         4         5 variable statistic
1   1999-01-01 00:00:00  5.40e-06  6.20e-06  8.000000e-06  1.20e-05  7.00e-06  evspsbl      MEAN
2   1999-01-01 01:00:00  3.90e-06  5.60e-06  7.300000e-06  1.25e-05  6.40e-06  evspsbl      MEAN
3   1999-01-01 02:00:00  3.20e-06  7.40e-06  5.700000e-06  6.80e-06  5.70e-06  evspsbl      MEAN
4   1999-01-01 03:00:00  8.30e-06  6.00e-06  4.200000e-06  1.80e-06  4.10e-06  evspsbl      MEAN
5   1999-01-01 04:00:00  5.10e-06 -1.00e-07  3.100000e-06  3.70e-06  3.00e-06  evspsbl      MEAN
6   1999-01-01 05:00:00  0.00e+00  1.90e-06  3.000000e-06  3.00e-06 -5.00e-07  evspsbl      MEAN
...
```

works, even though you are using more than one variable in this case. Since `mrso` isn't a variable used by our models, you can skip that. (We should also touch base on the variables necessary for the driver data, since not all variables in the Notaro dataset are used, they are [here](https://github.com/USGS-R/necsc-lake-modeling/blob/master/configs/ACCESS_config.yml#L3) but we renamed those for readability back then...).

_targets.R
```diff
@@ -24,29 +25,24 @@ targets_list <- list(
     split_lake_centroids(centroids_sf_rds)
   ),
 
-  # Convert the lake centroids to some kind of polygon for querying GDP
-  tar_target(
-    query_polys_sf,
-    centroids_to_poly(query_centroids_sf)
-  ),
-
   # Convert the sf polygons into a geoknife-ready format
```
hcorson-dosch, line 28: Just looks like you missed updating this comment here to refer to the centroids, rather than the lake polygons

hcorson-dosch: But then again we'll be tweaking this to instead use the grid cell centroids anyhow, so will change shortly

_targets.R
```diff
@@ -24,29 +25,24 @@ targets_list <- list(
     split_lake_centroids(centroids_sf_rds)
   ),
 
-  # Convert the lake centroids to some kind of polygon for querying GDP
-  tar_target(
-    query_polys_sf,
```
hcorson-dosch, line 29: Looks like this target is still used to build our test `query_map_png` below, FYI

hcorson-dosch (approved):
This all looks good to me, Lindsay. Wasn't sure if you wanted to test w/ `mrso` dropped to see if the `result()` parser works w/ multiple variables if that one isn't included?

PR #230, opened 2021-11-20
Wqp bbox

### LINKED ISSUE #229: Update WQP pull now that the bbox bug query in WQP services is fixed

jread-usgs:
Update to run WQP pulls now that the bbox query is fixed. 

Fixes #229


jread-usgs:
merge #218 first

1_crosswalk_fetch/src/fetch_wqp_sites.R
```diff
@@ -1,5 +1,7 @@
 fetch_wqp_sites <- function(out_ind, characteristicName, sf_ind, box_res, dummy, ...) {
 
+  sf::sf_use_s2(FALSE)
+  on.exit(sf::sf_use_s2(TRUE))
```
lindsayplatt, line 4: what is s2? I am reading the function help but it is not very illuminating.

lindsayplatt (commented):
Making sure I understand what is happening in this PR: this just kicked off the WQP site query after skipping it since Nov 2020 due to an issue with `whatWQPsites()` with `bBox` (related to [this commit?](https://github.com/USGS-R/lake-temperature-model-prep/commit/84aeefe044d3ac9d4f78c87c26c969acb248a9aa))

1_crosswalk_fetch/src/fetch_wqp_sites.R
jread-usgs: Good point! Will add comments

jread-usgs:
> Making sure I understand what is happening in this PR: this just kicked off the WQP site query after skipping it since Nov 2020 due to an issue with whatWQPsites() with bBox (related to this commit?)

the bbox param to the WQP webservices was failing and had to be avoided (see some discussion [here](https://github.com/USGS-R/dataRetrieval/issues/574) ) but the dev team recently addressed the bug in a recent release, so I kicked off this update, which I expected to be simple and mainly impact just .ind files and build .ymls. But I ran into the spatial issue noted above, so there was a small amount of code changed to get the build to complete.

Issue #236, opened 2021-12-01
Make the GCM grid in R


lindsayplatt:
Currently, our workflow used @hcorson-dosch's robust Python Notebook to generate the appropriate grid cell spatial files from the GCM grid on GDP. We want to avoid mixing and matching languages within this pipeline so that we don't introduce brittleness with environment setup. So, in #233, we just committed the resulting shapefiles and the Python notebook (`7_drivers_munge/in/Reconstruct_GCM_grid.ipynb`) that created them. 

This issue should try to use `st_make_grid()` to reconstruct the grid, as [this example](https://github.com/USGS-R/lake-surface-temperature-prep/blob/main/4_drivers_fetch/src/grid_utils.R#L13) does. If that doesn't work, we can build a multipolygon sf object instead (see [this comment](https://github.com/USGS-R/lake-temperature-model-prep/pull/233#issuecomment-984062211)). Once the new grid is built, we will no longer be relying on anything from the Python workflow and we can delete `7_drivers_munge/in/Reconstruct_GCM_grid.ipynb`.

hcorson-dosch:
I've got this workflow set up and running [here](https://github.com/hcorson-dosch/lake-temperature-model-prep/blob/gcm_driver_data_munge_pipeline/_targets.R#L23-L33), but once GDP is back online, we will need to add a target to directly download the netcdf file we're using to reconstruct the grid, instead of loading it in manually. Using `sf_make_grid()` does work, so long as you use the custom lambert conformal conic projection defined in the netCDF and set the offset using the projected coordinates from the netCDF.

jread-usgs:
Two thoughts here, since I hope it isn't needed to download the whole file locally to extract the metadata: 
- you can often reach NetCDF metadata (such as the spatial metadata you need to get the grid right) via the web without downloading the file. This is likely possible here when thredds is back online
- If you'd have to download the file to use it, you could 1) use the OPeNDAP subset algorithm (Lindsay used this for hurricane gifs) to just grab the entire spatial domain of the file, but only one timestep, making the file small, 2) instead consider hard-coding the projection and offset as variables. I am bringing up option 2 here because having a local file meant to represent the contents of a remote file can be error-prone (e.g., if the remote file changes projection due to being repacked/reprojected and/or optimized in some other way, we wouldn't know and would treat it the same if rebuilding the GCM targets) and hard-coding the information is likely about the same risk but is simpler.

lindsayplatt:
[...]

PR #238, opened 2021-12-02
Gcm driver data munge pipeline

hcorson-dosch:
* Per #236, switch to an R-based workflow to reconstruct the GCM grid, using the projection information and projection coordinates stored in the netCDF
* NOTE - I'm currently bringing in the GCM netCDF that I am using to reconstruct the grid manually, but once GDP is back up, we'll need to add a target to download that netCDF directly.
* This PR also deletes the `.ipynb` notebook and the zipped shapefiles that were used previously


7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -1,3 +1,105 @@
...
+# Construct grid tiles from a GCM netcdf
+# NOTE - sf_make_grid() can only  make square grid polygons
+# We're using tile_dim of 10 for now (tiles = 10 grid cells x 10 grid cells)
+# The grid is 110 cells wide by 85 high, and in this function we
+# get the number of rows/columns by floor(n_cells[high/wide]/tile_dim) so are rounding down.
+# With a tile_dim of 10, tiles won't cover full height of grid
+# but top 5 rows are fully outside CONUS, so we are dropping for now
+# rather than constructing two separate sf grids and merging
+construct_grid_tiles <- function(gcm_nc, tile_dim) {
+  gcm_data <- ncdf4::nc_open(gcm_nc)
```
lindsayplatt, line 59: A lot of the first part of this step is the same as is in `reconstruct_gcm_grid()`. Would it be possible to construct the `gcm_tiles_sfc` based on the output of `reconstruct_gcm_grid()` which would be the gcm grid? Otherwise, I would suggest grouping the common code into a shared function.

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -1,3 +1,105 @@
...
+# Reconstruct GCM grid from a GCM netcdf
+reconstruct_gcm_grid <- function(gcm_nc) {
+  gcm_data <- ncdf4::nc_open(gcm_nc)
+
+  # pull projection information
+  gcm_crs <- pull_gcm_crs(gcm_data)
+
+  # set grid size (there is a 'grid_size_in_meters' attribute,
+  # but I can't seem to pull it with ncatt_get)
+  grid_size_m <- 25000
```
lindsayplatt, line 22: Some of these GCM pieces are already hard-coded. It seems like it would be easies to just hard-code in some of the grid configurations and skip downloading and extracting the netCDF attributes. Seems like you could get away with a yaml that had these components, was read in as a list target, and then passed in as arguments to `reconstruct_gcm_grid()` and `construct_grid_tiles()`.

```yaml
# Grid configs
gcm_crs: 
grid_size_m: 
x_min: 
y_min: 
n_cell_x: 
n_cell_y: 
```

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: This becomes irrelevant if you do the second comment about moving configs into a yaml

lindsayplatt (approved):
The tibble target for the GCM grid configuration looks good

lindsayplatt:
Fixes #236

PR #239, opened 2021-12-02
Wqp bbox

jread-usgs:
This is an update to #230 which runs the build all the way through and addressed a minor encoding issue related to parsing the Bull Shoals data on mac

Updated map coverage is
image


7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -55,9 +54,11 @@ parse_Bull_Shoals_Lake_DO_and_Temp <- function(inind, outind) {
 
 # Thanks https://stackoverflow.com/questions/34024654/reading-rdata-file-with-different-encoding
 fix_colname_encoding <- function(df, new_encoding = c("UTF-8", "latin1")) {
-  cnames <- colnames(df)
-  Encoding(cnames) <- new_encoding
-  colnames(df) <- cnames
+  if(Sys.info()[['sysname']] == "Windows") {
```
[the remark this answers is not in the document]
lindsayplatt: Can we add a link to the scipiper issue where you described the Mac differences? Just since further changes will likely be described over there.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -37,16 +36,16 @@ parse_Bull_Shoals_Lake_DO_and_Temp <- function(inind, outind) {
       dplyr::rename(temp = "Temp (°C)") %>%
 
       # Replace "Surface" depth and convert depths to numeric
-      mutate(depth = as.numeric(ifelse(depth == "Surface", 0, depth))) %>%
+      dplyr::mutate(depth = as.numeric(ifelse(depth == "Surface", 0, depth))) %>%
 
       # Convert depth from feet to meters
-      mutate(depth = depth * 0.3048) %>%
+      dplyr::mutate(depth = depth * 0.3048) %>%
```
lindsayplatt, line 42: I never know how necessary this syntax is? Doesn't hurt to be explicit but can also be much cleaner to not do this. Do you have thoughts on when it is vs is not necessary?

lindsayplatt (approved):

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
jread-usgs: here it wasn't necessary. I did this for two reasons: within the function there was inconsistent use of `dplyr::` (e.g., at prior line 28, mutate was `dplyr::mutate` but `dplyr::` was missing for these lines. Second, I was getting an error with no method "mutate" for "Date" or something like that so my initial thought was that the wrong mutate function was being used. It turned out that error was a red herring and the real issue was related to the special encoding we discussed in https://github.com/USGS-R/scipiper/issues/151
But I kept my edits in. 

I use that syntax whenever I think there is potential for function name collisions or user interpretation confusion (e.g., "where does that function come from?"). I also use it when I don't want to list a package in the `packages: -...` part of the yml. Reason being, including a package there forces users to install it even if they aren't running that particular function (due to the shared cache) and in rare cases, those installs are a pain or are OS-specific, although we've mostly worked away from OS-specific issues here.

jread-usgs:
Yes. See https://github.com/USGS-R/scipiper/issues/151 discussion

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
lindsayplatt: Thanks for explaining!

PR #240, opened 2021-12-03
cache ind/build clean-up

jread-usgs:
This is a cache clean-up PR meant to resolve the false builds that were triggered in the coop section due to a gdrive file being out of sync with the ind and build files in the upstream repo. 

One of the challenges with shared cache pipelines in `scipiper` is that we need to keep the ind files and build files current, which often means committing changes ASAP and creating a PR soon when other people are working on similar stages of the pipeline at the same time as you or are working on parts of the pipeline downstream (but impacted by your changes) from you. 

To explain what was happening here: A file was changed on google drive because a target was built and part of that build checks if the file changed and if it did, it replaces the googld drive file with the new one. Then another person built stages of the pipeline and got file hash values that differed from what was expected in the build status file, which tells remake the target is out of date. But things don't end up updating in this case, and the system keeps assuming that file is out of date. To get past it, I did a force rebuild of the specific target, which deletes the file (the .ind) _and_ the status file (build/status/*.yml) and then rebuilds. This will also re-sync the file on google drive, and then the build becomes happy. 

All users of this repo will need this change otherwise they will see mysterious rebuilds that don't seem connected to any changes. @hcorson-dosch @lindsayplatt


lindsayplatt:
Since @padilla410 is also using and `scmake()`-ing this repo, want to tag her here. @jread-usgs should we talk about how to use `dry_put` during lengthy local dev?

lindsayplatt (approved):

lindsayplatt:
@hcorson-dosch I'll be sure to pull this update from `main` into our `gcm_driver_data_munge_pipeline`, too. @jread-usgs that shouldn't impact any of the builds because we aren't running `scmake()` with that piece yet and are just working on the targets side, but just note that we were keeping that effort in a separate branch on the canonical until we feel like it is ready.

padilla410:
@lindsayplatt, @jread-usgs and I briefly discussed this at the end of the day yesterday. Based on the level of completed so far on my piece of the pipeline, I'm just going to blow away my existing  local and remote repo and clone the repo from the canonical version (no lost time here).

I expect to be back into this on Monday afternoon.

PR #241, opened 2021-12-03
Reconcile recent build cache and ind with GCM munge branch

lindsayplatt:
Pull in the recent changes from `main` to this branch so that we aren't too far out of sink. We are not actually building `scmake()` right now with this pipeline, but good to keep this pipeline up-to-date with the other. Especially following #241


lindsayplatt:
All I did was `git pull upstream main` while on the `gcm_driver_data_munge_pipeline` branch. No additional changes or builds were made. Merging myself since these changes have already been reviewed in their own PRs.

jread-usgs:
Just a word of caution for having long-running branches on a scipiper pipeline that I think you are already aware of - 

Since there is no branch-specific sandbox for the remote data modified by code changes, in most situations it can introduce some headaches to use branches on `upstream` in these repos due to code/ind/build files changing in the different branches, but the remote data files having a single source of truth. You avoid this by either not building shared-cache targets on the branch (like you are doing) or using `dry_put = TRUE` in the scipiper config. Still a bit risky because sometimes `scmake` gets run out of habit :). I know you are aware of this, just spelling it out here because branching makes sense to use more commonly in most other code repos.

lindsayplatt:
Yes, 100% agree with those concerns. In this branch, we are only ever running `tar_make()`, so should not impact the shared cache or scipiper build files at all. This is a good reminder that when we are satisfied with the pipeline, we should merge with `main` AND THEN run the `scmake()` part to integrate the output of our side pipeline with the big scipiper one.

@hcorson-dosch tagging you here for awareness about this plan and a reminder about being cautious with how to use this branch while developing.

Issue #231, opened 2021-11-22
Make adding new coop data and parsers less brittle


jread-usgs:
- [x] ~need to update dummy arg when you add parsers~ fixed by #242 
- [x] ~download-and-push files are going into what we thought was a manual file transfer `in` folder (e.g., [here](https://github.com/USGS-R/lake-temperature-model-prep/blob/main/6_temp_coop_fetch.yml#L29)) and no check for duplicates~ fixed in #249
- [x] ~dummy arg update needed to check file list on gdrive~ fixed in #249
- [x] ~building `7a_temp_coop_munge/log/7a_temp_coop_munge_tasks.ind` doesn't trigger an update to the downloads of `6_temp_coop_fetch/log/6_temp_coop_fetch_tasks.ind` automatically (it isn't wired in as depends/input)~ fixed in #249 

Issue to capture discussion on this topic. 

There are a lot of small steps you need to remember to get through builds. Some targets need force builds (e.g.., parser inventory and gdrive file inventory)

Taking this example from an old `scipiper` discussion on how to make a target "always stale" in case it would help:

```r
target_default: all

packages:
  - sysfonts
  - showtext 

sources:
  - example_utils.R

targets:
  all:
    depends:
      - set_font
      - image.png

  set_font:
    command: set_fonts(I('Abel'), I('abel'),
      trigger_file = 'always_stale_time.txt')

 


  image.png:
    command: plot_things(target_name,
      trigger_file = 'always_stale_time.txt')
```
example_utils.R
```r
make_stale <- function(file){
  cat(file = file, paste0(format(Sys.time(), '%m/%d/%y %H:%M:%S '), sample(10000, 1)))
}

 

set_fonts <- function(..., trigger_file){
  make_stale(trigger_file)

 

  font_add_google(...)
  return(TRUE)
}

 
plot_things <- function(fileout, trigger_file){
  make_stale(trigger_file)

 


  png(filename = fileout)
  par(family = 'abel')
  showtext::showtext_begin()
  plot(cars)
  text(15, 100, 'this is a thing I plotted', cex = 2)
  showtext::showtext_end()
  dev.off()
}
```

jread-usgs:
I think there are a few brittle spots here, but one simpler one is the parser file list target in 7a_temp_coop_mung.yml:

```yaml
  7a_temp_coop_munge/tmp/parser_files.yml:
    command: list_coop_files(target_name,
[...]

PR #242, opened 2021-12-03
Create an "always stale" target that generates the hash table of parser functions

jread-usgs:
This is a partial for #231 , and focuses on solving a single problematic part of the pull: there is a target that generates the hash table of parser functions, but required manual intervention to trigger an update (the prior way was to edit the `dummy` string). This was generating issues, because if you made an edit to a parser function or added a new one, you wouldn't necessarily send those changes on to the parsing of the files. You'd discover this when you got an error because there was a cooperator-added file that didn't have a parser ("but I added that file!") and then realize you need to go back and change the  dummy argument. The worse case was when you mad a bug fix to a parser and thought you propagated the changes with `scmake()` but the changes weren't recognized because they are effectively hidden from `remake` (the parser files are referenced by a directory of files, not the individual files in the recipe). 

This is a `scipiper`-specific fix which does introduce some new codes and concepts even though I think we're doing most of the new stages and refactors with `targets`. But I found this issue to be too error-prone to have it keep biting us, so this could be a solution moving forward or could be replaced if the coop section of this pipeline is refactored with `targets` due to the stage's continued use/development and the other brittle components that still exist and are or will be documented in #231

What this solution does is force `7a_temp_coop_munge/tmp/parser_files.yml` and `coop_parsers` to both be stale on every single build. These are quick targets to generate, so this doesn't cost much. If neither of these targets actually change after rebuild, the result of the task table isn't rebuilt too, so that works as expected: 

image


lib/src/generic_utils.R
```diff
@@ -0,0 +1,11 @@
+
+#' write a unique value to a text file when this function is called.
+#' Replaces existing text in the file
+#'
+#' @param filepath a text filepath that can be written to
+#'
+#' @details this function is used to keep a dependency stale always, assuming this function is called whenever
+#' that file is referenced and there is at least a paired target use of the file.
+make_file_stale <- function(filepath){
+  cat(file = filepath, paste0(format(Sys.time(), '%m/%d/%y %H:%M:%S; random ID: '), sample(1E6, 1)))
```
limnoliver, line 10: What's the purpose of the random ID (as opposed to just MM-DD-YYYY H:M:S) ?

limnoliver (approved):
Good solution to this problem!

lib/src/generic_utils.R
jread-usgs: Thanks - I added a commit with more documentation explaining this. The random number is there because we want
to write something unique to the file, and there is a potential that this `make_file_stale()` function could be called several times in one second. Without the random ID added to it, the calls mad within the same second won't modify the file uniquely, and then it won't make the other targets stale.

PR #246, opened 2021-12-07
Use s2 false

### LINKED ISSUE #234: Add sf_use_s2(FALSE) to polygon overlap to avoid spherical coords issue with NHD polygons in sf > v1.0
Haven't tested, but seems this code snippet:
```r
sf::sf_use_s2(FALSE)
on.exit(sf::sf_use_s2(TRUE))
```

is needed [here](https://github.com/USGS-R/lake-temperature-model-prep/blob/master/2_crosswalk_munge/src/spatial_crosswalks.R#L61) if that function is going to cause an error. This was done for the points-in-poly function [here](https://github.com/USGS-R/lake-temperature-model-prep/pull/230/files#diff-c11bd7f597c7cb331c75bfe3d66ef9f7242c1842dde12e45e4f4afa9304509af) already, but it is expected to cause issues in the poly-to-poly overlap too.

jread-usgs:
fixes #234

```r
  sf::sf_use_s2(FALSE)
  on.exit(sf::sf_use_s2(TRUE))
```

lots of rebuilds triggered by this `sf`/PROJ change. Mostly ind/build commits here.


lindsayplatt (approved):
Wondering if we need to also add this during [intersections of lakes & grid cells](https://github.com/USGS-R/lake-temperature-model-prep/blob/gcm_driver_data_munge_pipeline/7_drivers_munge/src/GCM_driver_utils.R#L128) and also [grid cells & tiles](https://github.com/USGS-R/lake-temperature-model-prep/blob/gcm_driver_data_munge_pipeline/7_drivers_munge/src/GCM_driver_utils.R#L110)?

jread-usgs:
It isn't likely needed elsewhere when the spatial data are simpler than NHD. NHD has some polygons that break the spherical change, and we'd seemingly need to do a lot of things to repair them if we wanted to use S2, so the simpler solution here is just to revert to the older way of representing them for this part of the analysis.

Issue #248, opened 2021-12-09
Troubleshooting initial pipeline build


padilla410:
Using this issue to move an email/Team conversation into the repo. This issue is related to #219 and will be used to track any problems executing the full coop data pipeline through the `scmake("8_viz")` step.

This issue will be closed at the same time as #219 and will be considered closed after a successful run of each of the following steps:
* Get the data (`scmake("6_temp_coop_fetch")`)
* Munge  (`scmake("7a_temp_coop_munge")`)
* Combine  (`scmake("7b_temp_merge")`)
* Plot (`scmake("8_viz")`)

After adding `parse_navico_data.R` I was able to run `scmake("6_temp_coop_fetch")`. Running the second step (`scmake("7a_temp_coop_munge")`) resulted in the following errors:
![image](https://user-images.githubusercontent.com/15007294/145480860-3723faae-5f83-40b9-bce1-42387a18da4f.png)
![image](https://user-images.githubusercontent.com/15007294/145480873-5f3615ef-3451-43e8-ab86-e5c79186d564.png)
![image](https://user-images.githubusercontent.com/15007294/145480878-9d61efd1-3f55-4d7f-8a02-2c982e1ceab9.png)

padilla410:
@jread-usgs email reply (lightly edited to connect links):

Julie, I am cc'ing Lindsay because she ran into a similar situation with the same file. 

You are hitting this error because there are multiple files on google drive in the same folder with the same name ("long_term_ntl.csv") and there is a check in scipiper for this condition that throws an error (see [Scipiper PR 100](https://github.com/USGS-R/scipiper/pull/100)). Lindsay and I discussed this a bit ago when she ran into it but we kind of brushed it off as a weird mystery that we worked past by deleting the second file copy of the file on google drive. But you've created several versions of this file on drive and it is the same file Lindsay had issue with, so there is something up (see below)!

![image](https://user-images.githubusercontent.com/15007294/145481750-58cdc31c-bbc3-45a2-aed2-d7835fbabe47.png)


[...]

PR #249, opened 2021-12-10
Brittle coop

jread-usgs:
Getting this PR up because it has ind/build changes. ~Will add more context later this am.~ But is tied to #231 and things @padilla410 discovered in #248 

#### update
The focus of this PR is to make working on new temperature data sources more developer-friendly, by clearing out several of the gotchas in target connections and triggers that in hindsight, explain a number of mystery builds from the past. 

- I've pulled in changes from Julie's open PR so that we're not colliding on indicator file changes. I will keep up with updates or merge of that PR if things change and get them into this PR too. Pulling in changes from an open PR is generally not a practice we'd do, but in scipiper/shared cache repos, it is sometimes necessary. 
- I moved the "forever stale" target from #242 to 1_fetch so it could be used by other steps easily
- I added the stale trigger file to two targets in 6_coop, which include indexing the files on google (was triggered by a dummy string in the past) and the `coop_wants` target that relies on the index ind. It is necessary to pair targets to use the trigger file, otherwise it doesn't work correctly (this explains why `coop_wants` uses it too)
- since the google drive index will happen with every build that depends on it now and that call to google is expensive and we could get API-limited if accessing too often(?), I put in a time check on the indicator for the file index (see below for `freshened: 2021-12-10 05:44:22 -0600`) and it won't re-index google drive unless the last build was less than `trigger_wait` minutes ago (I set this to 30). This seems a little questionable, but I feel confident it is a major improvement over what was happening in the past
- the NTL targets were removed, which was to address #248. Those are in the `in` folder anyhow and we don't need to re-download them. I did go ahead and update one of the files since a newer version was available, so that explains the file name and parser change in here
- I linked the downloaded file index to the 7a_ munge task table, since those weren't connected before. Likely another source of confusing double builds that happened in the past. 
- Added `NULL` as default on the trigger file targets, so leaving out that arg will skip the trigger stale step in case you want to avoid this when developing.


6_temp_coop_fetch/out/coop_all_files.rds.ind
```diff
@@ -1,2 +1,3 @@
-hash: 59795c7efcc7526f0462cdac07698f21
+hash: 7566d6ac6bfe10e47a9a3a90065e827e
+freshened: 2021-12-10 05:44:22 -0600
```
jread-usgs, line 2: going to `scdel()` this and rename from .ind to .tind (time-indicator) to avoid confusion with the extra information

jread-usgs: Never mind, that isn't possible because the various `scipiper` functions check that it has a valid name for an indicator file.

limnoliver:
Nice fixes to some questionable choices I made 🙈! Just a few comments -- I still don't understand why you need the always stale trigger in both targets? (but I see your note, so trust you've thought it through)

7a_temp_coop_munge/src/data_parsers/parse_wilter_files.R
```diff
@@ -15,7 +15,7 @@ parse_mendota_daily_buoy <- function(inind, outind) {
   sc_indicate(ind_file = outind, data_file = outfile)
 }
 
-parse_long_term_ntl <- function(inind, outind) {
+parse_ntl29_v10_0 <- function(inind, outind) {
```
jread-usgs, line 18: will impact #247 open PR if it re-pulls google drive files

6_temp_coop_fetch.yml
```diff
@@ -17,46 +17,36 @@ targets:
...
+
       - 6_temp_coop_fetch/log/6_temp_coop_fetch_tasks.ind
 
   # -- download LTER files and push to GD -- #
   coop_file_upload_location:
     command: as_id(I('1dutCiFEOoRObXLn6n3BRKhZDDAijK65Q'))
 
-  6_temp_coop_fetch/downloads/long_term_ntl.csv.ind:
-    command: download_and_indicate(in_ind = target_name,
-             url = I("https://lter.limnology.wisc.edu/sites/default/files/data/ntl29_v5.csv"),
```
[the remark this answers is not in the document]
jread-usgs: It was originally going into that `in` automatically too, since in the `download_and_indicate()` function:
```r
googledrive::drive_upload(media = local_file, path = push_location)
```
Where `push_location` is the google drive location for the `in` folder. So this was going into `downloads` _and_ into `in`. `in` is where it became part of the coop parsers and I think the `downloads` folder is ignored by this pipeline. 

> Is the download URL still being documented somewhere?
yes, I added an explainer file for this target.

6_temp_coop_fetch/log/6_temp_coop_fetch_tasks.ind
```diff
@@ -97,15 +98,15 @@
...
 6_temp_coop_fetch/in/Winnie_Temp_2010.xlsx.ind: 5167ff0a9232e5e829dfa0f16e537e85
 6_temp_coop_fetch/in/Winnie_Temp_2011.xlsx.ind: ea2ddeb745b1735d29ff8045a4d2b86c
+6_temp_coop_fetch/in/Winnie_temp_loggers_2013.xlsx.ind: 76392c97ee043eba66f7522e020875b5
 6_temp_coop_fetch/in/Winnie_temp_2014.xlsx.ind: 2074717df365d6c1253c813230aac402
-6_temp_coop_fetch/in/DNRdatarequest_Secchi_DO_and_Temp_1083_2016_AllLakes_JL.xlsx.ind: 22c8cc6fc3af2311c2d90edf9940fedb
 6_temp_coop_fetch/in/Winnie_temp_logger_output_2015.xlsx.ind: 2af920209709ea42afb3ace38e0f0bdc
 6_temp_coop_fetch/in/Winnie_temp_logger_2017.xlsx.ind: 3c93b9db2f6fae33db885a32f38302f5
+6_temp_coop_fetch/in/Winnie_water_temp_2012.xlsx.ind: 9116b7df6d69d55519e843e9f5336a18
 6_temp_coop_fetch/in/Winnie_temp_2_2009.xlsx.ind: c19ce966b9c5a6072c0cf1d1516dd2f2
+6_temp_coop_fetch/in/DNRdatarequest_Secchi_DO_and_Temp_1083_2016_AllLakes_JL.xlsx.ind: 22c8cc6fc3af2311c2d90edf9940fedb
```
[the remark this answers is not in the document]
jread-usgs: I created [this scipiper issue](https://github.com/USGS-R/scipiper/issues/153) and I think the ordering that matters here is in `sc_indicate` instead of the result from `drive_ls`, since this `.ind` file is the result of the task table and not of the GD inventory function. 

But yes, you are seeing the re-ordering, not a new hash. It is distracting to see diffs when the hashes aren't changing. I'm with ya

6_temp_coop_fetch.yml
```diff
@@ -17,46 +17,36 @@ targets:
...
-    command: files_in_drive(out_ind = target_name, gd_path = coop_file_upload_location, dummy_date = I('2021-12-01'))
+    command: files_in_drive(out_ind = target_name,
+      gd_path = coop_file_upload_location,
+      trigger_file = '1_crosswalk_fetch/out/always_stale_time.txt',
+      trigger_wait = 30)
 
   coop_wants:
-    command: filter_coop_all('6_temp_coop_fetch/out/coop_all_files.rds.ind')
+    command: filter_coop_all('6_temp_coop_fetch/out/coop_all_files.rds.ind',
+      trigger_file = '1_crosswalk_fetch/out/always_stale_time.txt')
```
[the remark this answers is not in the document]
jread-usgs: As far as I can tell, this is related to the way remake is tracking build changes while things are executing. If we only have one target in the dependency chain that has the trigger file, it oddly is not considered stale the second build (even though you would expect it to, since it has secretly modified the trigger_file as a side-effect of the build). I can't remember exactly what I determined here - perhaps if you don't have two trigger_file-associated targets, you only get rebuilds every other time, which is frustrating and confusing. So my solution is to always pair them when used as this is reliable 100% of the time.

6_temp_coop_fetch/out/coop_all_files.rds.ind
jread-usgs: Modified this with an updated commit

PR #251, opened 2021-12-10
GCM driver data pipeline: export the lake_cell_xwalk as a CSV file

lindsayplatt:
@hcorson-dosch this is for https://github.com/USGS-R/lake-temperature-process-models/issues/3. It took me way too many commits to do this 🤣


hcorson-dosch (approved):
Perfect, thanks!

Issue #255, opened 2021-12-14
File contents changes on google drive do not trigger re-downloads


jread-usgs:
This was noted in #249 and in the review [comment](https://github.com/USGS-R/lake-temperature-model-prep/pull/249#discussion_r766714536) and some work was set up to make tracking this possible, including exposing the GD file hash in the `6_temp_coop_fetch/out/coop_all_files.rds` file and the `coop_wants` object. But as of now, only the filenames go into the task tables, so downloads won't be triggered unless the filename has changed. 

This could be addressed by adding the `I({hash})` as a dummy arg to the task table for each download or some other solution that uses the hash.

PR #256, opened 2021-12-14
whoops! didn't get the tibble format all the way through

jread-usgs:
Related to prep work for #255


padilla410 (approved):
Reviewed in a screenshare with @jread-usgs

Issue #259, opened 2021-12-14
Navico data: expand `navico_nhdhr_xwalk.rds` crosswalk


padilla410:
The existing crosswalk for mapping the `Navico_ID` field in the Navico temperature data set to `site_id` in `navico_nhdhr_xwalk.rds` is incomplete.

The current raw  temperature dataset (`Waterbody_Temperatures_by_State.rds`) has 2,786 unique entries:
```r
> # unique `Navico_ID` in raw data
> length(unique(navico_raw$Navico_ID))
[1] 2786

> # unique `site_id` in crosswalk
> length(unique(navico_xwalk$site_id))
[1] 658
```

Only 10% of the unique waterbodies in the raw dataset have matching values in the crosswalk:
```r
> # join the raw Navico data to the cross walk and count unique `Navico_ID`
> left_join(navico_raw, navico_xwalk) %>% 
+   filter(!is.na(site_id)) %>% 
+   pull(Navico_ID) %>% 
+   unique() %>% 
+   length() 
Joining, by = "Navico_ID"
[1] 296
```

Finally, while working through this issue I also discovered that there is a non-unique relationship between `Navico_ID` and `site_id` in `navico_nhdhr_xwalk.rds`. This issue is fully documented in #257, but here is a snippet of what is happening:
```r
> # join the raw Navico data to the cross walk and count unique `site_id`
> left_join(navico_raw, navico_xwalk) %>% 
+   filter(!is.na(site_id)) %>% 
+   pull(site_id) %>% 
+   unique() %>% 
+   length()
Joining, by = "Navico_ID"
[1] 294

> # identify non-unique `site_id`s
> left_join(navico_raw, navico_xwalk) %>% 
+   filter(!is.na(site_id)) %>% 
+   group_by(site_id, Navico_ID) %>% 
+   tally() %>% 
+   group_by(site_id) %>% 
+   tally() %>% 
+   filter(n > 1)
Joining, by = "Navico_ID"
# A tibble: 2 x 2
  site_id             n
  <chr>           <int>
1 nhdhr_155637149     2
2 nhdhr_155641743     2
```

padilla410:
I think it depends on your definition of "the same lake" that you've been using for the model. 

I looked at this two ways:

[...]

Issue #257, opened 2021-12-14
Navico data: non-unique relationship between `site_id` and `Navico_ID` in nhdr crosswalk


padilla410:
I'm not 100% sure if this is a true issue or not, but it did strike me as odd: there is a non-unique relationship between `Navico_ID` and `site_id` in the `lake-temperature-model-prep > 2_crosswalk_munge > out > navico_nhdhr_xwalk.rds` on google drive.

Here is a length comparison between the two fields:
```r
> navico_xwalk <- readRDS('data/2021-12-06/2_crosswalk_munge_out/navico_nhdhr_xwalk.rds') # update with your local path
> length(unique(navico_xwalk$Navico_ID))
[1] 667
> length(unique(navico_xwalk$site_id))
[1] 658
```

A count summary of all the dups:
```r
> navico_xwalk %>% group_by(site_id) %>% tally() %>% filter(n > 1)
# A tibble: 7 x 2
  site_id                                          n
  <chr>                                        <int>
1 nhdhr_{0BB28A37-D665-4735-A2DA-5C08014F9EEF}     2
2 nhdhr_{d7eeb7a2-c7ce-45bd-b329-7038adea7656}     2
3 nhdhr_132544104                                  3
4 nhdhr_155637149                                  2
5 nhdhr_155638724                                  2
6 nhdhr_155641743                                  2
7 nhdhr_155642288                                  3
```

A detailed look at one of the duplicates:
```r
> navico_xwalk %>% filter(site_id == "nhdhr_132544104") %>% select(site_id, Navico_ID, WaterbodyName, State, County)
# A tibble: 3 x 5
  site_id         Navico_ID      WaterbodyName    State County          
  <chr>           <chr>          <chr>            <chr> <chr>           
1 nhdhr_132544104 Navico_1226897 Lower Gar Lake   Iowa  Dickinson County
2 nhdhr_132544104 Navico_1226898 Minnewashta Lake Iowa  Dickinson County
3 nhdhr_132544104 Navico_1253574 Upper Gar Lake   Iowa  Dickinson County
```

Based on a quick view of the area on google maps, it looks like this is a spatial resolution issue: Upper Gar Lake, Minnewashta Lake, and Lower Gar Lake are part of a chain of lakes:
![image](https://user-images.githubusercontent.com/15007294/146059568-885c711b-108f-4149-87ff-f7abd1196b90.png)

jread-usgs:
[...]

PR #260, opened 2021-12-15
Expanding Navico crosswalk to sites with temperatures but no depths

jread-usgs:
This is to unblock @padilla410 

I made a decision to not create a separate (and redundant) 1_fetch target for the `Waterbody_Temperatures_by_State` sites because these are already contained in the coop task table. 
~I think normally I would've kept the depths separate as a 1_fetch target (basically, as-is prior to this PR) and the temperature sites as a different 1_fetch target, but the fact that the primary purpose of the `Waterbody_Temperatures_by_State` file is in the coop section and we're using a slightly different caching concept there, I combined the two.~

**update**
I decided to modify my PR to separate the Navico depths and temperatures in 1_fetch and instead create a new merge target in 2_ for all points (`2_crosswalk_munge/out/navico_lakes_sf.rds.ind`) and then use that to build the crosswalk for Navico w/ both the depth locations and water temp locations. 

This means that `merge_navico_lakes()` is going to try to `gd_get()` the .xlsx file using the coop tasktable. If that file doesn't exist, this function will fail, but I provided an error message that explains it. 

I did an `scmake()` of `4_params_munge`, which formats the depths to later go into `7_config_merge`, but didn't build all the way there on this PR, as I think there was still some investigation going on as to how many new sites this dataset adds and I didn't want to overwrite our current GD data without knowing what we were tallying.


padilla410:
All of this looks good to me and I have documented answers to the one question I had (in the review).

I was able to successfully run the following commands without error:
* `scmake("2_crosswalk_munge")`
* `scmake("4_params_munge")`

I also pulled down an updated copy of `navico_nhdhr_xwalk.rds` to review the updates with respect to #259 and #257:

```r
> # unique `Navico_ID` in raw data
> length(unique(navico_raw$Navico_ID))
[1] 2786

> # unique `Navico_ID` in raw data left joined to crosswalk
> length(unique(navico_nhd_complete$Navico_ID))
[1] 1533

> # unique `site_id` in raw data left joined to crosswalk
> length(unique(navico_nhd_complete$site_id))
[1] 1510
```

With Jordan's update ~55%  of the unique waterbodies in the raw dataset have matching values in the crosswalk. Clearly there are still duplicate `site_id`s in the crosswalk (result in 1510 uniquw `site_id`s), but we've decided that #257 is a `wontfix`

padilla410 (approved):
All of this looks good to me and I have documented answers to the one question I had (in the review).

I was able to successfully run the following commands without error:
* `scmake("2_crosswalk_munge")`
* `scmake("4_params_munge")`

I also pulled down an updated copy of `navico_nhdhr_xwalk.rds` to review the updates with respect to #259 and #257:

```r
> # unique `Navico_ID` in raw data
> length(unique(navico_raw$Navico_ID))
[1] 2786

> # unique `Navico_ID` in raw data left joined to crosswalk
> length(unique(navico_nhd_complete$Navico_ID))
[1] 1533

> # unique `site_id` in raw data left joined to crosswalk
> length(unique(navico_nhd_complete$site_id))
[1] 1510
```

With Jordan's update ~55%  of the unique waterbodies in the raw dataset have matching values in the crosswalk. Clearly there are still duplicate `site_id`s in the crosswalk (result in 1510 unique `site_id`s), but we've decided that #257 is a `wontfix`

PR #210, opened 2021-06-29
Add navico

hcorson-dosch:
Added Navico depth data to pipeline -- closes #207 

* Add data to Drive
  - [x] Add Navico dataset to Google Drive - `1_crosswalk_fetch/in/`
  - [x] Add getter for Navico dataset in `getters.yml`
  - [x] Run `scipiper::gd_confirm_posted()` on the raw file (with `.ind` extension)
* Changes in `1_crosswalk_fetch.yml`
  - [x] Add in intermediate indicator file for sf object `.rds` target
  - [x] Add .rds.ind to getters.yml 
* Changes in `2_crosswalk_munge`
  - [x] Create NHD_Navico crosswalk ind file
  - [x] Add crosswalk to `gettters.yml` as `2_crosswalk_munge/out/` file
* Changes to `4_params_munge`
  - [x] Munge Navico depth data using NHD_Navico crosswalk - make new `4_params_munge/out/` ind file -- use 'depth' in the target name
  - [x] Add munged data ind file to `getters.yml`
* Changes to `7_config_merge.yml`
  - [x] Add new Navico munged depth data in `4_params_munge/out/` to argument list for `7_config_merge/out/nml_H_A_values.rds.ind` - make first of 'depth' args
* Changes to `8_viz.yml`
  - [x] None, but `scmake('8_viz')` should complete all needed builds

PR #263, opened 2021-12-20
Adding Navico data and parser

### LINKED ISSUE #219: Bring in Navico temperature data
Incorporate the [Navico lake temperature data](https://docs.google.com/spreadsheets/d/1bw_Kett2A6WcbYQNjB2jYFof2EkLtrKX/edit?usp=sharing&ouid=100995528200180469387&rtpof=true&sd=true) into the pipeline.

Initial notes:
* The data are from lakes in 11 states (AR, IL, IN, IA, MI, MN, MO, ND, OH, SD, WI)
* The temperature fields are average, maximum, and minimum water temp
* The data includes averages over varying timescales (all time, each year, each month, each day) that will need to be filtered out - there are separate columns for Year, Month, Day, Hour, which are noted as 'All' when data is averaged over the corresponding period
* No depth is included - we're assuming these are surface temperatures taken with a thermistor sensor attached to a sonar unit on a boat
* The data are in separate state-specific tabs within an Excel spreadsheet
* Waterbody latitude and longitude are provided, as are waterbody name and the Navico MapWaterbodyID. We should hopefully be able to use this last ID  (renamed to Navico ID in this pipeline) to match to the previously incorporated Navico max and mean depth data.
* From a quick glance, looks like there may sometimes be >1 site per lake, named differently by waterbody name and (I think) MapWaterbodyID (renamed to Navico ID in this pipeline)

padilla410:
This pull request is the second half of PR #247 and successfully runs the following components of the pipeline:

1. Fetch data ('scmake("6_temp_coop_fetch")`)
2. Munge data into correct format (`scmake("7a_temp_coop_munge")`)
3. Combine into one data set (`scmake("7b_temp_merge")`)
4. Plot to evaluate data changes (`scmake("8_viz")`) (<-- recognizing that this step alone would result in a complete build)

The main work here is in `7a_temp-coop_munge/src/crosswalk_fxns.R` and a few tweaks in `7a_temp-coop_munge/src/data_parsers/parse_navico_data.R` to get the pipeline to recognize the Navico data crosswalk in `2_crosswalk_munge/out/`

Completing the pipeline run results in 17,167 records in `7a_coop_munge/out/all_coop_dat_linked.feather`. All other datasets except two, appear to be identical (based on counts). The two datasets that are not identical are `long_term_ntl.rds` and `ntl29_v10_0.rds`:
image
_(Note: in this case "`n_old` refers to `all_coop_dat_linked.feather` from 2021-12-08 and `n_new` refers to `all_coop_dat_linked.feather` from 2021-12-17)_

I assume the discrepancy in the table above is due to the NTL download issues described in issue #248 and resolved in PR #249. 

When comparing `all_coop_dat_linked.feather` from 2021-12-08 to the same file from 2021-12-17, there are 895 **new** waterbodies that have corresponding NHD IDs. This seems reasonable based on the fact that there were 2,786 waterbodies in the original data set and 1,510 with corresponding NHD IDs. That being said, when comparing the html screenshot below to the other most recent HTML screenshot (from PR #239), it looks like we only gain 3 waterbodies (see PDGL Lake (n = _xxx_)).

Closes #219 

Screenshot of updated map coverage (`8_viz/out/lakes_summary_fig.html`):
image


padilla410:
Update: 1,508 waterbodies _not_ 1,510. based on the following snippet: 

```r
x <- read_feather(sc_retrieve('7a_temp_coop_munge/out/all_coop_dat_linked.feather.ind'))

x %>% 
  filter(source == '7a_temp_coop_munge/tmp/Waterbody_Temperatures_by_State.rds') %>%
  pull(site_id) %>% 
  unique() %>% 
  length()
  length(unique(site))
```

lindsayplatt:
Attempting to dig through this in chunks. 

>  The two datasets that are not identical are long_term_ntl.rds and ntl29_v10_0.rds ... I assume the discrepancy in the table above is due to the NTL download issues described in issue #248 and resolved in PR #249.

Correct! In addition to fixing the "always stale" issues, #249 deleted the previous version of this NTL data (called `long_term_ntl.rds`) from `6_temp_coop_fetch/in` and replaced with the newest available data from [North Temperate Lakes LTER](https://lter.limnology.wisc.edu/) (called `ntl29_v10_0.csv`). The newer one has more data, so that checks out to me 👍  

Checking on the counts next and then will scan through all the code!

lindsayplatt:
I am struggling to get the right combo to narrow down which 3 lakes, but I am comfortable with the fact that you aren't seeing a huge uptick in the count of waterbodies that are model-able. This Navico data largely replaced other water temperature data that we already had (presumably it is of higher quality). In `7b_temp_merge.yml`, all of the temp data is combined into one giant feather file, `7b_temp_merge/out/merged_temp_data_alltimes.feather.ind`, using `merge_temp_data()`. THEN, the dataset is resolved into one feather file `7b_temp_merge/out/merged_temp_data_daily.feather.ind` with a single observation for each lake, date, and depth using `reduce_temp_data()`.

If we compare the before/after of reducing the giant temp data into one observation per lake/date/depth, we can start to see how the Navico data _replaces_ a lot of data rather than _adds_ more to our pipeline. So this is largely an increase in _quality_ not _quantity_.

Here is how I started to explore this:

```r
library(scipiper)
library(arrow)
library(tidyverse)
options(scipiper.dry_put = TRUE)

source_navico <- '7a_temp_coop_munge/tmp/Waterbody_Temperatures_by_State.rds'

# All temperatures merged contain the Navico data (17,132)
all_temp_data <- read_feather(sc_retrieve("7b_temp_merge/out/merged_temp_data_alltimes.feather.ind"))
all_temp_navico <- all_temp_data %>% filter(source == source_navico)
nrow(all_temp_navico)

# After the target the resolves multiple unique obs per lake/date/depth, there are 8,048 rows of Navico data
# in the final dataset. This confirms that it is, in fact, getting added but my assumption is that it
# is replacing other data, not adding new data
resolved_temp_data <- read_feather(sc_retrieve("7b_temp_merge/out/merged_temp_data_daily.feather.ind"))
resolved_temp_navico <- resolved_temp_data %>% filter(source == source_navico)
nrow(resolved_temp_navico)

# Keep only the Navico sites & days/depths that overlap with our recently added data
navico_overlap <- resolved_temp_navico %>% select(site_id, date, depth)
all_temp_shared_navico <- left_join(navico_overlap, all_temp_data)

# Here is the data that was replaced by the Navico data
overlap_summary <- all_temp_shared_navico %>% 
  group_by(site_id, date, depth) %>% 
  mutate(count = n()) %>% 
  ungroup() %>% 
  filter(count > 1)

head(overlap_summary)
```

You can see in the overlap summary how a single lake/date/depth combo can have multiple sources (the first two rows show both Navico AND another source of data, but there could be more).
```r
# A tibble: 6 x 11
  site_id              date       depth time  timezone  temp source_id     source_site_id   source                  month count
  <chr>                <date>     <dbl> <chr> <chr>    <dbl> <chr>         <chr>            <chr>                   <dbl> <int>
1 nhdhr_{0BB28A37-D66~ 2020-05-29     0 14:15 EST       14.9 LRBOI_WQX-LM~ LRBOI_WQX-LMich  wqp_LRBOI_WQX-LMich         5     2
2 nhdhr_{0BB28A37-D66~ 2020-05-29     0 14:00 NA        16.5 Navico        NA               7a_temp_coop_munge/tmp~     5     2
3 nhdhr_120018754      2019-07-14     0 13:00 CDT       26.9 MNPCA-61-007~ MNPCA-61-0072-0~ wqp_MNPCA-61-0072-00-2~     7     2
4 nhdhr_120018754      2019-07-14     0 15:00 NA        27.4 Navico        NA               7a_temp_coop_munge/tmp~     7     2
5 nhdhr_120018981      2019-09-15     0 09:30 CDT       18.5 MNPCA-11-020~ MNPCA-11-0203-0~ wqp_MNPCA-11-0203-01-2~     9     6
6 nhdhr_120018981      2019-09-15     0 18:00 CDT       16.7 MNPCA-11-020~ MNPCA-11-0203-0~ wqp_MNPCA-11-0203-01-2~     9     6
```

Note that while the Navico has 17k+ observations, many of them were for the same lake/date/depth but at different _times_ of the day. We are resolving into _daily_ observations so we lose a bunch of quantity in this way, too. For example:

```r
overlap_summary %>% filter(count == 14)
```

```
# A tibble: 14 x 11
   site_id        date       depth time  timezone  temp source_id     source_site_id source                         month count
   <chr>          <date>     <dbl> <chr> <chr>    <dbl> <chr>         <chr>          <chr>                          <dbl> <int>
 1 nhdhr_74926427 2018-05-20     0 00:00 NA        16.8 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 2 nhdhr_74926427 2018-05-20     0 01:00 NA        15.8 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 3 nhdhr_74926427 2018-05-20     0 02:00 NA        15.9 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 4 nhdhr_74926427 2018-05-20     0 03:00 NA        16.0 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 5 nhdhr_74926427 2018-05-20     0 13:00 NA        14.7 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 6 nhdhr_74926427 2018-05-20     0 14:00 NA        15.3 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 7 nhdhr_74926427 2018-05-20     0 15:00 NA        15.5 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 8 nhdhr_74926427 2018-05-20     0 16:00 NA        16.3 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
 9 nhdhr_74926427 2018-05-20     0 17:00 NA        17.1 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
10 nhdhr_74926427 2018-05-20     0 18:00 NA        18.0 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
11 nhdhr_74926427 2018-05-20     0 19:00 NA        17.9 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
12 nhdhr_74926427 2018-05-20     0 21:00 NA        20.8 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
13 nhdhr_74926427 2018-05-20     0 22:00 NA        18.0 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
14 nhdhr_74926427 2018-05-20     0 23:00 NA        18.9 Navico_915714 NA             7a_temp_coop_munge/tmp/Waterb~     5    14
```

7a_temp_coop_munge.yml
```diff
@@ -46,7 +46,7 @@ targets:
   7a_temp_coop_munge/log/7a_temp_coop_munge_tasks.ind:
     command: loop_tasks(task_plan = coop_munge_taskplan,
       task_makefile = '7a_temp_coop_munge_tasks.yml',
-      num_tries = 3)
+      num_tries = 1)
```
lindsayplatt, line 49: I think we may want to leave `num_tries = 3` unless there was a specific reason we dropped it lower?

lindsayplatt (approved):
This all looks correct to me! Just spent some time thinking through those counts and have question on the change to `num_tries`.

7a_temp_coop_munge.yml
padilla410: @jread-usgs and I changed this value from 1 to 3 while troubleshooting so that we could save some wait time. We decided to commit the change because `num_tries` is typically intended to query a flaky web service multiple times without breaking the pipeline. In the case of this specific pipeline, it's just repeating the error 3 times - if it fails once it will definitely fail thrice.

padilla410:
@lindsayplatt thank you for the detailed review! I was unaware that this dataset was intended to replace a lot of the data currently in the pipeline.

padilla410:
> ```r
> options(scipiper.dry_put = TRUE)
> ```

Thank you for this example! We'd been having conversations around how to use `dry_put` but I did not understand how to implement. A lot of the troubleshooting and review I did was in a small, offline `Rproj` to avoid triggering an update of the data on google drive.

lindsayplatt:
> @lindsayplatt thank you for the detailed review! I was unaware that this dataset was intended to replace a lot of the data currently in the pipeline.

I am not sure if we knew how much would be new vs replacing other data when we started this, but maybe I am out of the loop on Navico. I am remembering not too much of a difference when [the crosswalk was added here](https://github.com/USGS-R/lake-temperature-model-prep/pull/210#issuecomment-872305573).

7a_temp_coop_munge.yml
jread-usgs: I suggested 1 because this isn't a brittle function call that fails sometimes and works other times (like a flakey web request). My understanding is if any one of these fails once, they will fail the other two times, hence no need for a retry here.

Issue #258, opened 2021-12-14
GCM pipeline: scale up to MN


lindsayplatt:
Considerations for scaling up to MN (and beyond 🚀 👩‍🚀)

- How long does the current GCM pipeline take since we are setup to use all lake centroids? This likely means that _most_ of the downscaled GCM cells will have at least one lake in them. Does it make since to continue with our current pattern or should we try to download the GDP results as NetCDF files and setup the pipeline to operate using those? Some thoughts on this from @jread-usgs in [this comment](https://github.com/USGS-R/lake-temperature-model-prep/issues/252#issuecomment-992542631) and in [this one](https://github.com/USGS-R/lake-temperature-model-prep/issues/252#issuecomment-993672205). Ultimately, we should probably limit the lake centroids to ones that are "model-able" (define as having some depth values?) to limit the downscaled GCM data we are pulling.
- If we do continue with the current download approach, explore and iterate on the tile size we should use to optimize download times and memory usage.
- Currently (or very soon with an upcoming PR), our data transformations to get GLM variables from GCM ones includes a couple of pivoting steps and then a series of `tidyverse` commands to `mutate`, `select`, etc. This (especially pivoting) could be very slow. Consider switching to matrix operations to increase the speed.

lindsayplatt:
Pivoting: consider moving to `data.table::dcast()` and `data.table::melt()` instead of `tidyr::pivot_*` as ways to optimize the code quickly. Just need to convert the tibble to a data.table. Consider the following pattern for going from long to wide:

```r
tibble() %>% setDT() %>% dcast(...) %>% as_tibble()
```

lindsayplatt:
Scaled up the time period to be the full time period for each of the 3 (1980-1999, 2040-2060, and 2080-2099) + added in all of the GCMs. Still only doing this for 5 lakes (and only 4 different cells). Download times are not looking great (16 hours in total ranging from 13 to 40 min), but the munge steps (even with all the pivoting) are taking just over 2 minutes in total.

```r
library(targets)
library(dplyr)

summarize_target_build <- function(branch_names) {
  target_info <- tar_meta(all_of(branch_names)) %>% 
    unnest(path) %>% 
    mutate(size_mb = bytes*0.000001,
           duration_min = seconds/60) %>% 
    select(path, time, duration_min, size_mb)
  
  ## PRINT OUT SOME RESULTS
  
  dur_min <- round(min(target_info$duration_min), digits = 3)
[...]

Issue #252, opened 2021-12-11
GCM pipeline - NetCDF formatting of output


lindsayplatt:
An initial implementation of NetCDF output was created, but there are a few concerns that are causing us to rethink the approach. 

1. Built from scratch and so we are deviating from established standards, which means we would need to edit when we go release. Instead, we should use [functions from ncdfgeom](https://github.com/USGS-R/ncdfgeom/blob/master/R/write_timeseries_dsg.R) to write out the NetCDF file for timeseries spatial data (example use of these functions [here in lakesurf-data-release](https://github.com/jread-usgs/lakesurf-data-release/blob/main/src/ncdf_utils.R#L66-L76)).
2. Looping through each variable and each cell to populate the data in the NetCDF file (see [this commit](https://github.com/lindsayplatt/lake-temperature-model-prep/commit/66a9c42cce5885bc5469aad7b0f8e2c1243a7aa9) only on Lindsay's fork here), will be slow and isn't necessary.

We also need to consider how to extract the information, as well, since we want to create individual files for each GCM, time period, and cell to use in our [lake-temperature-process-models](https://github.com/USGS-R/lake-temperature-process-models) repo (this splitting will happen over in that pipeline).

Going into this process, we have `feather` files for each tile (groups of cells) that look like this:

```
> read_feather("7_drivers_munge/tmp/7_GCM_ACCESS_1980_1999_tile37_daily.feather")
# A tibble: 14 x 9
   time        cell Shortwave Longwave AirTemp  RelHum       Rain       Snow WindSpeed
   <date>     <dbl>     <dbl>    <dbl>   <dbl>   <dbl>      <dbl>      <dbl>     <dbl>
 1 1980-01-02  3996      57.1     57.1  -5.24  0.00231 0          0               3.57
 2 1980-01-03  3996      48.1     48.1  -6.73  0.00182 0          0               2.98
 3 1980-01-04  3996      45.6     45.6  -2.45  0.00296 0          0               3.64
 4 1980-01-05  3996      40.3     40.3   0.437 0.00375 0.00000625 0               2.64
 5 1980-01-06  3996      68.6     68.6  -5.09  0.00258 0.0000675  0.000454        4.53
 6 1980-01-07  3996      91.9     91.9 -12.9   0.00110 0          0.00000417      2.58
 7 1980-01-08  3996      88.7     88.7  -7.65  0.00174 0          0               4.65
[...]

PR #266, opened 2021-12-21
Munge GCM downloads into GLM-ready variables

### LINKED ISSUE #264: Expand beyond test time period for downscaled GCMs
Do this for just the 5 test lakes and send resulting feather files to @hcorson-dosch so she can practicing running GLM on the full time period.

lindsayplatt:
In addition to the munging, this adds

- a temporary way to get all the GCM data as feather files to skip NetCDF files & my original attempt at making a NetCDF file (see #252) and unblock Hayley (those files were zipped up and manually uploaded [here in Sharepoint](https://doimspp.sharepoint.com/:u:/r/sites/IIDDStaff/Shared%20Documents/Project%20-%20Modeling%20Lakes/GLM/lake-temperature-process-models/ProcessModel_1PrepTmp_files_20211221.zip?csf=1&web=1&e=ggNuqx) for now)
- expands the time ranges to use the full time period (fixes #264)

Putting this up here now but just want it to sit for now. Need to do the following before it can be reviewed:
- [x] Delete NetCDF attempt (or actually solve by doing #252)
- [x] add `retry()` to GCM download code
- [ ] ~add some basic optimization to the pivoting code~ (waiting to do this later given mixed results. See https://github.com/USGS-R/lake-temperature-model-prep/issues/258#issuecomment-1006063738)

### DIFF
--- a/7_drivers_munge/src/GCM_driver_nc_utils.R
+++ b/7_drivers_munge/src/GCM_driver_nc_utils.R
@@ -13,149 +13,15 @@
 generate_gcm_nc <- function(nc_file, gcm_raw_files, dim_time_input, dim_cell_input,
                             vars_info, global_att, overwrite = TRUE) {
 
-  # Setup the NetCDF file with appropriate dimensions and attributes
-  build_gcm_nc_skeleton(nc_file, dim_time_input, dim_cell_input, vars_info, global_att, overwrite = overwrite)
+  # Use `ncdfgeom` to write out the NetCDF DSG file
+  # See example at https://github.com/jread-usgs/lakesurf-data-release/blob/main/src/ncdf_utils.R#L66-L76
+  # And more discussion in https://github.com/USGS-R/lake-temperature-model-prep/issues/252
+  # ncdfgeom::write_timeseries_dsg(nc_file)
 
-  # Populate the NetCDF file with data from each of the tiles.
-  # Opens and then closes the `nc_file` each loop iteration.
-  for(tile_fn in gcm_raw_files) {
-    arrow::read_feather(tile_fn) %>%
-      push_gcm_nc_data(nc_file, var_names = vars_info$var_name)
-  }
+  # TODO: remove this step which just creates an empty file
+  # with a .nc extension as a placeholder
+  writeLines("PLACEHOLDER", nc_file)
 
   return(nc_file)
 }
-
-
-#' @title Creates a blank NetCDF file for storing GCM driver data
-#' @description Creates a NetCDF file with appropriate dimensions and NA values
-#' to be filled in with real values where available in `push_gcm_nc_data()`. This
-#' function was adapted from functions shared by J. Zwart on 12/03/2021. See more at
-#' https://code.usgs.gov/wma/wp/forecast-pgdl-da/-/blob/main/2_forecast_prep/src/nc_utils.R
-#' @param nc_file netcdf file name output
-#' @param dim_time_input vector of GCM driver data dates
-#' @param dim_cell_input vector of all GCM driver grid cells (whether or not data was pulled)
-#' @param vars_info variables and descriptions to store in NetCDF
-#' @param global_att global attribute description for the data in this NetCDF file (e.g. notaro_ACCESS_1980_1999)
-#' @param overwrite T/F if the nc file should be overwritten if it exists already
-build_gcm_nc_skeleton <- function(nc_file, dim_time_input, dim_cell_input, vars_info, global_att, overwrite = TRUE) {
-
-  ##### Set dimensions; should always be [time, cell] #####
-
-  # Setup time dimension
-  times <- as.integer(seq(0, length(dim_time_input) - 1, 1)) # days since dim_time_input[1]
-  time_dim <- ncdim_def("time",
-                        units = sprintf('days since %s', dim_time_input[1]),
-                        longname = 'time',
-                        vals = times)
-
-  # Setup grid cell dimension
-  cells_dim <- ncdim_def("gridcells",
-                         units = "",
-                         longname = 'GCM grid cell number from lake-temperature-model-prep',
-                         vals = dim_cell_input)
-
-  # Check that units are valid
-  udunits2::ud.is.parseable(time_dim$units)
-  udunits2::ud.is.parseable(cells_dim$units)
-
-  ##### Define variables #####
-
-  fillvalue <- NA # fill for missing data
-
-  def_list <- list()
-
-  n_vars <- nrow(vars_info)
-  # loop through observation variables and define name, units, precision
-  for(i in 1:n_vars){
-    cur_var <- vars_info$var_name[i]
-
-    def_list[[i]] <- ncvar_def(name =  vars_info$var_name[i],
-                               units = vars_info$units[i],
-                               dim = list(time_dim, cells_dim),
-                               missval = fillvalue,
-                               longname = vars_info$longname[i],
-                               prec = vars_info$precision[i])
-  }
-
-  ##### Write out the file #####
-
-  if(file.exists(nc_file)){
-    if(overwrite){
-      file.remove(nc_file)
-      ncout <- nc_create(nc_file, def_list, force_v4 = T)
-    }else{stop('cannot overwrite nc output file')}
-  }else{ncout <- nc_create(nc_file, def_list, force_v4 = T)}
-
-
-  ##### Add global file metadata #####
-  ncatt_put(nc = ncout,
-            varid = 0,
-            attname = "file_description",
-            attval = global_att,
-            prec =  "text")
-
-  nc_close(ncout)
-  return(nc_file)
-
-}
-
-#' @title insert GCM data into an existing NetCDF file
-#' @description Using the NetCDF file created from `build_gcm_nc_skeleton()`, this
-#' function will populate the file with the actual data.
-#' @param gcm_df data frame with the GCM data to insert. Assumes this data has a
-#' date column, a column called `variable` containing values that can be matched to
-#' `var_names`, and a column for each of the grid cell's data named by the cell number.
-#' This argument is first so that the function can be piped to pass in the data.frame.
-#' @param nc_file file path to the netcdf file
-#' @param var_names vector of variable names to loop through
-push_gcm_nc_data <- function(gcm_df, nc_file, var_names){
-
-  # Open the NetCDF file
-  ncout <- nc_open(nc_file, write = T)
-
-  # Cycle through each of the variables (outer `for` loop) & then the grid cells
-  # containing data (inner `for` loop). For each, extract a vector of containing
-  # data for the current cell and variable and then insert into the appropriate
-  # dimensions within the NetCDF file.
-  for(cur_var in var_names){
-
-    # Extract the current variable's info from the nc data
-    cur_var_att <- ncout$var[[cur_var]]
-    varsize <- cur_var_att$varsize
-    n_dims <- cur_var_att$ndims
-
-    # Order of the GCM dimensions as created in `build_gcm_nc_skeleton()`; [time_dim, cells_dim]
-    dim_time_pos <- 1
-    dim_cells_pos <- 2
-
-    # Extract GCM values for the current variable from the bigger GCM data frame
-    cur_var_df <- dplyr::filter(gcm_df, variable == cur_var)
-
-    # Cycle through only cells that are in the `gcm_df`rather than all 9000+ grid cells for each variable
-    # and each file. To find the cells that have data, force column names to be numeric and then only keep
-    # ones that succeed (so this assumes that the only numeric column names are the grid cell columns, which
-    # seems reasonable). As of 12/7, non-cell colnames are "date", "variable", "statistic", and "units"
-    available_cells <- na.omit(as.numeric(names(gcm_df)))
-
-    for(cur_cell in available_cells){
-
-      # Define where to start & how many values to add for each dimension [time_dim, cells_dim]
-      start <- c(1, cur_cell)
-      count <- c(varsize[dim_time_pos], 1) # Use all of the slots for the time dimension, but only one for the cell dimension
-
-      # Extract GCM values for the current cell from the corresponding column in the GCM data frame
-      cur_cell_vals <- cur_var_df[[as.character(cur_cell)]]
-
-      ncvar_put(nc = ncout,
-                varid = cur_var,
-                vals = cur_cell_vals,
-                start = start,
-                count = count)
-    }
-  }
-
-  nc_close(ncout)
-}
-
 
--- a/7_drivers_munge/src/GCM_driver_nc_utils.R
+++ b/7_drivers_munge/src/GCM_driver_nc_utils.R
@@ -1,0 +1,168 @@
+# Munging driver data from geoknife to create NetCDF for each GCM.
+
+# # Example of extracting one cell's timeseries data
+# x <- nc_open("7_drivers_munge/out/7_GCM_ACCESS.nc")
+# cell_i <- 4650
+# xx <- ncvar_get(x, "evspsbl", start = c(1,cell_i), count = c(-1,1))
+
+#' @title Creates a NetCDF file with GCM driver data
+#' @description For each of the GCMs, create a single NetCDF file containing all the
+#' variables and all the grid cells.
+#' @param nc_file netcdf file name output
+#' @param gcm_raw_files vector of feather file paths, one for each tile with data for the current GCM.
+#' @param dim_time_input vector of GCM driver data dates
+#' @param dim_cell_input vector of all GCM driver grid cells (whether or not data was pulled)
+#' @param vars_info variables and descriptions to store in NetCDF
+#' @param global_att global attribute description for the observations (e.g. notaro_ACCESS_1980_1999)
+#' @param overwrite T/F if the nc file should be overwritten if it exists already
+generate_gcm_nc <- function(nc_file, gcm_raw_files, dim_time_input, dim_cell_input,
+                            vars_info, global_att, overwrite = TRUE) {
+
+  # Setup the NetCDF file with appropriate dimensions and attributes
+  build_gcm_nc_skeleton(nc_file, dim_time_input, dim_cell_input, vars_info, global_att, overwrite = overwrite)
+
+  # Populate the NetCDF file with data from each of the tiles.
+  # Opens and then closes the `nc_file` each loop iteration.
+  for(tile_fn in gcm_raw_files) {
+    arrow::read_feather(tile_fn) %>%
+      push_gcm_nc_data(nc_file, var_names = vars_info$var_name)
+  }
+
+  return(nc_file)
+}
+
+
+#' @title Creates a blank NetCDF file for storing GCM driver data
+#' @description Creates a NetCDF file with appropriate dimensions and NA values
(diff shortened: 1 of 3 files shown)


7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -242,26 +245,125 @@ download_gcm_data <- function(out_file_template, query_geom,
...
+                 names_to = "cell", values_to = "val") %>%
+    mutate(cell = as.numeric(cell)) %>%
+
+    # Now pivot wider to makes each variable a column for easy, readable munging
+    # Also, removes `units` and `statistic` columns which are specific to each variable
+    pivot_wider(id_cols = c("DateTime", "cell"), names_from = variable, values_from = val) %>%
+
+    # Unit conversions to get GLM-ready variables from GDP ones
+    mutate(AirTemp = tas - 273.15, # Convert GDP air temperature (tas) from Kelvin to Celcius
+           Rain = pr # TODO: Convert GDP precipitation (pr) from kg/m2/s to meters/hour
```
lindsayplatt, line 322: @jread-usgs interested in what your thoughts are for this variable.

lindsayplatt:
@jread-usgs looking for your review specifically on the `munge_to_glm()` function but feel free to comment on other aspects. @hcorson-dosch looking for your review on my `targets` code. Note that I deleted all of my old NetCDF munging code, but left the top-level function, `generate_gcm_nc()`, there (and all its params) but it just saves a blank .nc file for now (this will get filled out in #252

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
   ),
 
   ##### Munge GDP output into NetCDF files that will feed into GLM #####
-  # TODO - munge data
+
+  # Need to munge GCM variables into useable GLM variables
+  tar_target(glm_vars_info,
+             tibble(
+               # TODO: MISSING LONGWAVE
+               var_name = c("Rain", "AirTemp", "Snow", "RelHum", "Shortwave", "Longwave", "WindSpeed"),
```
hcorson-dosch, line 156: Looks like the order of this column doesn't match the order of the other columns - "AirTemp" and "Snow" are swapped

7_drivers_munge/src/GCM_driver_utils.R
jread-usgs: Ahh, I looked at Hayley's issue #15 before this so expect some duplicate info here :)

For this conversion, you can assume the density of water is 1000kg/m3 and use that to calculate the flux in the GLM units.

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
+               longname = c("Total daily rainfall",
+                            "Total daily snowfall derived from precipitation and air temp",
+                            "Average daily near surface air temperature",
+                            "Average daily percent relative humidity",
+                            "Average daily surface downward shortwave flux in air",
+                            "LONGWAVE EXPLANATION",
+                            "Average daily windspeed derived from anemometric zonal and anenometric meridional wind components"),
+               units = c("meters", "meters", "degrees Celcius", "", "???", "???", "???"),
```
hcorson-dosch, line 164: For GLM, we need these units:
| GLM driver input | [Units needed by GLM](https://gmd.copernicus.org/preprints/gmd-2017-257/gmd-2017-257-manuscript-version3.pdf) - see table at p59 |
| ----------- | ----------- | 
| Rain | m/day |
| Snow | m/day |
| AirTemp | degrees celsius |
| RelHum | percent ( 0 - 100) |
| Shortwave      | W/m2  |
| Longwave   | W/m2 |
| WindSpeed | m/s |

And see Jordan's notes about conversions [here](https://github.com/USGS-R/lake-temperature-process-models/issues/15#issuecomment-1006084022)

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
+  tar_target(gcm_names, c('ACCESS', 'GFDL', 'CNRM', 'IPSL', 'MRI', 'MIROC')),
+  tar_target(gcm_dates_df,
+             tibble(
+               projection_period = c('1980_1999', '2040_2059', '2080_2099'),
+               start_datetime = c('1980-01-01 00:00:00', '2040-01-01 00:00:00', '2080-01-01 00:00:00'),
+               # Include midnight on the final day of the time period
+               end_datetime = c('2000-01-01 00:00:00', '2060-01-01 00:00:00', '2100-01-01 00:00:00')
+             )),
+  # Gathered manually from https://cida.usgs.gov/thredds/ncss/notaro_GFDL_2040_2059/dataset.html
+  tar_target(gcm_query_vars, c("pr", "tas", "qas", "rsns", "uas", "vas")), # missing longwave radiation
```
hcorson-dosch, line 129: Just for our records, might be helpful to include what each of these corresponds to in a comment, since the abbreviations aren't super helpful.
image

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
 
   # Download data from GDP for each tile & GCM name combination.
```
hcorson-dosch, line 131: Looks like this missed being updated to note that you're now crossing to download data for each tile & GCM name & projection period combination

7_drivers_munge/src/GCM_driver_utils.R
hcorson-dosch: And note we'll actually need units of meters/day, not meters/hour. I just did this to check the GLM results, and that conversion comes out to `m/day = kg m-2 s-1 * 86.4`

Here's how I got to that so that you can double check it and make sure it seems right:

kg m-2 s-1 = density of water (kg/m3) * rate of rainfall (m/s)

Since we want the rate of rainfall, we divide the GDP value in kg m-2 s-1 by the density of water (1000 kg/m3), and then multiply by 86400 (60 secs * 60 min * 24 hrs) to convert the resultant rate (m/s) to m/day. So that results in a conversion factor of 86.4.

_targets.R
hcorson-dosch: Oh and looks like we might be able to use that `ps` surface pressure variable to [calculate RelHum](https://github.com/USGS-R/lake-temperature-model-prep/blob/main/7_drivers_munge/src/GLM_driver_utils.R#L151) from the `qas` specific humidity variable.

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
-    format = "file"
+    pattern = cross(query_cells_centroids_list_by_tile, gcm_names, gcm_dates_df),
+    format = "file",
+    error = "continue"
   ),
 
   ##### Munge GDP output into NetCDF files that will feed into GLM #####
-  # TODO - munge data
+
+  # Need to munge GCM variables into useable GLM variables
```
hcorson-dosch, line 152: Maybe this comment should go with the next target, since the info provided by this target isn't used in the actual munging?

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -242,26 +245,125 @@ download_gcm_data <- function(out_file_template, query_geom,
...
+  tar_meta(all_of(dynamic_branch_names), c("path", "data")) %>%
+    # There is just one path per branch, so remove the `list` format from this column
+    unnest(path)
+}
+
+#' @title Summarize hourly output from geoknife to daily
+#' @description the final GCM driver data will need to be daily, but geoknife
+#' returns hourly values. This step summarizes the data into daily values. It
+#' creates a file with the exact same name, except that the "_raw" part of the
+#' `in_file` filepath is replaced with "_daily".
```
hcorson-dosch, line 305: Might be good to add the conversion steps to this function description

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
+  # TODO: delete once we finish the NetCDF DSG build (see issue #252)
+  tar_target(
+    out_skipnc_feather, {
+      out_dir <- "7_drivers_munge/out_skipnc"
+      purrr::map(gcm_data_daily_feather, function(fn, gcm_names, gcm_dates_df) {
+        gcm_name <- str_extract(fn, paste(gcm_names, collapse="|"))
+        gcm_time_period <- str_extract(fn, paste(gcm_dates_df$projection_period, collapse="|"))
+
+        read_feather(fn) %>%
+          split(.$cell) %>%
```
hcorson-dosch, line 186: oh, interesting! I've never seen this `split()` function before

7_drivers_munge/src/GCM_driver_utils.R
jread-usgs: Nice work, Hayley. One thing to add here is that the _GDP_ values aren't always going to be in the same units (different data sources on GDP have different units) but we can assume the GCMs have the same units. An important check we can use on this conversions is to verify the units column in the GDP result is what we're expecting. That is one good reason to bring along that optional column output. That way if we point this to a slightly different dataset/variable in the future, we'd fail early and we'd know why if the units differed.

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -242,26 +245,125 @@ download_gcm_data <- function(out_file_template, query_geom,
...
+#' @param in_file filepath to a feather file containing the hourly geoknife data
+munge_to_glm <- function(in_file) {
```
jread-usgs, line 307: might want to call this `munge_notaro_to_glm` since this function is specific to a particular set of variables and unites. 

In the future, might want to pull out the generic munging components into a separate function and keep the munge prep work that is specific to a particular source in a unique function. (e.g., could use the same generic for NLDAS and Notaro GCMs, but unique prep functions for both). But probably not important now.

jread-usgs (commented):
I focused on the `munge_to_glm` function.

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
+      }, gcm_names, gcm_dates_df)
+      return(out_dir)
+    },
+    format = "file"
+  ),
+
+  # Group daily feather files by GCM to map over and include
+  # branch file hashes to trigger rebuilds for groups as needed
+  tar_target(gcm_data_daily_feather_group_by_gcm,
+              build_branch_file_hash_table(names(gcm_data_daily_feather)) %>%
```
hcorson-dosch, line 204: This is a nice solution for tracking file dependencies while getting the grouping we'd like.

_targets.R
```diff
@@ -115,35 +117,128 @@ targets_list <- list(
...
+               group_by(gcm_name) %>%
+               tar_group(),
+             iteration = "group"),
+
+  # Create a single vector of all the time period's days to use as the NetCDF time dim
+  tar_target(
+    gcm_dates_all,
+    purrr::pmap(gcm_dates_df, function(projection_period, start_datetime, end_datetime) {
+      seq(as.Date(start_datetime), as.Date(end_datetime), by = 1)
+    }) %>% do.call("c", args = .)
```
hcorson-dosch, line 216: This line just vectorizes the list that is returned by `purrr::pmap`?

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -0,0 +1,27 @@
...
+
+#' @title Creates a NetCDF file with GCM driver data
+#' @description For each of the GCMs, create a single NetCDF file containing all the
+#' variables and all the grid cells.
+#' @param nc_file netcdf file name output
+#' @param gcm_raw_files vector of feather file paths, one for each tile with data for the current GCM.
+#' @param dim_time_input vector of GCM driver data dates
+#' @param dim_cell_input vector of all GCM driver grid cells (whether or not data was pulled)
+#' @param vars_info variables and descriptions to store in NetCDF
+#' @param global_att global attribute description for the observations (e.g. notaro_ACCESS_1980_1999)
```
hcorson-dosch, line 11: Looks like you may have tweaked the format of this global attribute description since providing this example.

lindsayplatt:
Commenting just to say that I am seeing all these comments and starting to work through them! Thanks for these very helpful reviews @jread-usgs and @hcorson-dosch

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -242,26 +245,125 @@ download_gcm_data <- function(out_file_template, query_geom,
...
+      # The default knife algorithm is appropriate
     )
-    #TODO: should we specify the `knife`??
-  )
-  wait(gcm_job)
+    wait(gcm_job)
+    gcm_job_status <- check(gcm_job)$statusType
+  },
+  # Check the value of the last thing in the expr above (denoted
+  # by the ".") to decide if you should retry or not
```
hcorson-dosch, line 267: Thanks for adding such good comments - really helps to make clear here what is being checked

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -242,26 +245,125 @@ download_gcm_data <- function(out_file_template, query_geom,
...
+    ) %>%
+
+    # Simply rename GDP variables into GLM variables
+    mutate(RelHum = qas,
+           Shortwave = rsns,
+           Longwave = rsns # TODO: find longwave from GDP
+    ) %>%
+
+    # Calculate GLM variables using other existing variables
+    mutate(Snow = ifelse(AirTemp < 0, Rain*10, 0), # When air temp is below freezing, any precip should be considered snow
```
hcorson-dosch, line 332: Just out of curiosity, @jread-usgs and @lindsayplatt, why is the snow value set to 10 times the rain when the temp is below 0? I see that's done [here](https://github.com/USGS-R/lake-temperature-model-prep/blob/main/7_drivers_munge/src/GLM_driver_utils.R#L153) too. Is 10 just a 'best guess' factor being used to account for the much lower density of snow?

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -242,26 +245,125 @@ download_gcm_data <- function(out_file_template, query_geom,
...
+           RelHum,
+           WindSpeed,
+           Rain,
+           Snow
+    ) %>%
+
+    # Convert from hourly to daily data
+    group_by(time, cell) %>%
+    # TODO: should we `na.rm = TRUE`?
+    summarize(across(.cols = -WindSpeed, .fns = ~ mean(.x, na.rm = FALSE)),
```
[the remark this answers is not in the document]
hcorson-dosch: Agreed! This is a really clean approach.

_targets.R
hcorson-dosch: (just wanting to make sure I'm understanding it correctly, as I haven't seen `do.call()` previously)

hcorson-dosch (commented):
Okay - done with my review! I looked over the `targets` code as well as the `munge_glm()` function. Thanks for all of the detailed comments and function descriptions, Lindsay. Those really helped. I think this is looking great, and will be good to merge in once the unit conversions are finalized.

7_drivers_munge/src/GCM_driver_utils.R
jread-usgs: yes, 10 assumes snow depth is 10x the water equivalent. This varies depending on the type of snow, but is a good starting point. 

I also mentioned to Lindsay that we have the option of leaving the snow column blank, which then has GLM make this calculation internally.

hcorson-dosch: okay, thanks!

7_drivers_munge/src/GCM_driver_nc_utils.R
lindsayplatt: Oh yes. I should probably just delete all of this documentation since I anticipate arguments passed to the function will change immensely with #252 but this runs in the pipeline, so was keeping for now.

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: Hmmm what would you say are the "generic munging components" here? Feels like most are Notaro specific

7_drivers_munge/src/GCM_driver_utils.R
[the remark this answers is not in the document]
lindsayplatt: This gets a bit muddied by the precip need - "meters/day". Wouldn't this need a `sum` of the hourly rate and not a `mean`?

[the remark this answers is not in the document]
lindsayplatt: I think I can just follow the pattern I did for `WindSpeed` where it has its own method defined.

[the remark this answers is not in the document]
hcorson-dosch: Since it's a rate, rather than the depth value that some models use, I think using mean() is fine, right @jread-usgs ?

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: Ahh OK, so we are getting the _rate_ into meters/day (even though we will have a different daily rate per hourly timestep after this `mutate()`, then we will average those daily rainfall rates to get a single daily rainfall rate per day in `summarize()` below. So, my comment about summing per day is not what we want.

_targets.R
lindsayplatt: Oops! Thank goodness for other eyes

_targets.R
lindsayplatt: It can be pretty handy! It is similar to a `group_by()` in that you specify a column to group rows of the data. Just nice here because you can do more within the next `purrr::map()` call than you would do with a `summarize()`

_targets.R
lindsayplatt: Ahh yes! That is exactly what it is doing. `purrr:pmap()` returned a list. `do.call()` is one way of collapsing lists. `unlist()` does something equivalent to this here, but unfortunately doesn't keep the `Date` class.

lindsayplatt:
Working my way through the following list after combing through your reviews:

- [ ] change fxn name to include "Notaro"
- [ ] add more description about the unit conversions in the munge fxn descr
- [ ] edit rain conversion to be meters/day
- [ ] edit relhum conversion (need to download additional variable from GDP)
- [ ] investigate conversion of datetimes to dates (check tz etc)
- [ ] see if tz fix means we are still losing day 1
- [ ] add variable descriptions
- [ ] update comment about downloading to mention each time period, too
- [ ] fix `glm_vars_info` to be correct order
- [ ] add appropriate units to `glm_vars_info`

7_drivers_munge/src/GCM_driver_utils.R
jread-usgs: I see generic as: going from hourly to daily, writing the files, and perhaps some of the variable changes. Probably good to just ignore this since I agree there isn't a clear divide and NLDAS isn't handled here currently anyhow.

7_drivers_munge/src/GCM_driver_utils.R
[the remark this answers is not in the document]
jread-usgs: yes, mean would be equivalent in this case. If it were m/hour though, we'd need to treat it differently

lindsayplatt:
Pausing for now (just relative humidity and tz stuff left) because the updated downscaled, debiased GCM data might make some of this moot. See #273

lindsayplatt:
Decided that merging this, but noting those two incomplete items - timezone & correct relative humidity conversions - is the best way to keep moving forward. Given the new GCMs are already daily and mostly in the units we need already, these two concerns likely won't be present anyways.

Issue #273, opened 2022-01-06
Switch to debiased GCMs


lindsayplatt:
David has those available now on THREDDS. Fixes the data but also adds longwave radiation, which will need to be a new variable added to our workflow.

- [x] Switch to downloading the new data (now daily and you can download 1980-2099 all at once)
- [x] Check to see if [these two issues](https://github.com/USGS-R/lake-temperature-model-prep/pull/266#issuecomment-1006799360) (timezone woes & relative humidity conversions) are still needed
- [x] Verify the grid projections are correct (see discussion [here](https://github.com/USGS-R/lake-temperature-model-prep/pull/276#discussion_r781372224)) (completed, read comments in this issue below)
- [x] Check that the dates (and missing Dec 31sts) are OK (see details in [this comment](https://github.com/USGS-R/lake-temperature-model-prep/pull/276#issuecomment-1013264491)) _Fixed by #299_ 
- [x] (moved to its own issue) #296
- [x] Munged feathers in target `glm_ready_gcm_data_feather` [sometimes have `NA`s and sometimes have `NaN`s](https://github.com/USGS-R/lake-temperature-model-prep/pull/281#issuecomment-1028343457). How come? This causes an issue with building our NetCDF file, so we added a fix that converts `NA`s to `NaN`s to get put into the NetCDF file.

lindsayplatt:
Here are where the new NC files are for this data. The variables available here are already formatted and in the units we would use for GLM, which makes the recent unit conversion effort a bit moot. See

* [WindSpeed (m/s)](http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias/files/debias_windspeed_mri_mid21_longerdaymet_ghcn_365d.nc.html)
* [Snow (cm) & air temperature (degC)](http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias/files/debias_snv_tas_mri_mid21_longerdaymet_ghcn_365d.nc.html)
* [Shortwave radiation](http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias/files/debias_rsds_mri_mid21_longerdaymet_ghcn_365d.nc.html)
* [Longwave radiation](http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias/files/debias_rsdl_mri_mid21_longerdaymet_ghcn_365d.nc.html)
[...]

Issue #271, opened 2022-01-04
Mega-Issue: Refactor Pipeline


padilla410:
Over the past few months a number of small issues have occurred when working in the pipeline that have made debugging difficult and will ultimately result in code that is difficult to maintain. 

The current issue is a _**work in progress**_. It represents my current understanding of what we would like to refactor based on a few meetings from 2022-01-04. Tagged assignees should edit this issue based on their current understanding of problems with the pipeline. Problem definition should include why a portion of the pipeline needs refactoring, notes on how to refactor (if you've given it some thought), and desired outcomes from the refactor (simplified code? An output table? etc). The Lakes Team will then have a meeting to see if we are all on the same page and to discuss timeline for completion. After that meeting, I will update/finalize this issue and start the refactoring process.

### Refactor crosswalks from `1_crosswalk_fetch/in` that will be used in `7a_temp_coop_munge`
**Problem:** Using the current pipeline configuration, each unique dataset munged in `7a_temp_coop_munge` needs a crosswalk in  `1_crosswalk_fetch\in`. This results in an ever expanding number of arguments being passed into [`crosswalk_coop_dat`](https://github.com/USGS-R/lake-temperature-model-prep/blob/main/7a_temp_coop_munge/src/crosswalk_fxns.R). The ever expanding list of arguments has two problems: 
* There is currently a lot of [repeated code within the function](https://github.com/USGS-R/lake-temperature-model-prep/blob/0c037ddce5bdab7a7eb3a6ad07c0837972bddcd1/7a_temp_coop_munge/src/crosswalk_fxns.R#L17-L25) which will ultimately result in code that is difficult to maintain into the future.
* A waterbody is not always unique to a dataset. For example, data for Bull Shoals Reservoir exists in at least three data sets (Navico #257, Univ of Missouri #269, and the citizen science Bull Shoals data set from PR #218). While not a problem per se, it again adds repetition to `crosswalk_coop_dat` (i.e., does the citizen science data set from Bull Shoals really need it's own crosswalk when it only contains one lake?)
Normalizing crosswalks across datasets may make it possible to cut down on arguments to `crosswalk_coop_dat`, minimize duplicate work, and produce a crosswalk that links cooperator  data to `nhdhr` IDs which is a known desired output.

**Key Outcome:**
* A final crosswalk that maps local ID values to `nhdhr` id values that can be returned to cooperators as a data release

[...]

PR #276, opened 2022-01-08
Code to switch to new GCMs & scale up to MN

lindsayplatt:
Scaling up to MN will require one additional step, which is to have a more robust way to subset our lakes to MN than what I did [here](https://github.com/USGS-R/lake-temperature-model-prep/commit/e6358c37f075d90f5c3fa2aba77787d3718998b8). I was able to test this with those 5 specific NML sites (see [this commit](https://github.com/USGS-R/lake-temperature-model-prep/commit/6f20b1ca67a0b761dc34218c4d45052d94995bd8)) and it worked (see the notes [here](https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1007856274)). Now I am running this for our new 

This addresses #273 and #258. Leaving as draft until my `tar_make(glm_ready_gcm_data_feather)` command finishes.

_Note that this pull request includes the results following a `git pull origin main`, resulting in so many commits and changes. Since we have kept the GCM pipeline in a separate branch, we did this step to get the updates from `main` into this branch. [This commit](https://github.com/USGS-R/lake-temperature-model-prep/pull/276/commits/55187c5db7096b9030c11c13130e37b71b6d0f27) is really the first commit of new changes related to the GCM work._


_targets.R
```diff
@@ -117,24 +117,24 @@ targets_list <- list(
...
+  tar_target(gcm_names, c('ACCESS', 'GFDL', 'CNRM', 'IPSL', 'MRI', 'MIROC5')),
   tar_target(gcm_dates_df,
              tibble(
-               projection_period = c('1980_1999', '2040_2059', '2080_2099'),
-               start_datetime = c('1980-01-01 00:00:00', '2040-01-01 00:00:00', '2080-01-01 00:00:00'),
+               projection_period = c('1980_2099'),
+               start_datetime = c('1980-01-01 00:00:00'),
                # Include midnight on the final day of the time period
-               end_datetime = c('2000-01-01 00:00:00', '2060-01-01 00:00:00', '2100-01-01 00:00:00')
+               end_datetime = c('2100-01-01 00:00:00')
```
jread-usgs, line 126: Does it actually go to 2100 or is the last date in 2099?

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -245,8 +236,13 @@ download_gcm_data <- function(out_file_template, query_geom,
...
 
   # Build query_url
-  query_url <- sprintf("https://cida.usgs.gov/thredds/dodsC/notaro_%s_%s",
-                       gcm_name, gcm_projection_period)
+  query_url <- sprintf(
+    "http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias_%s",
+    tolower(gcm_name))
+
+  # Stop now if you aren't on VPN
+  if(!RCurl::url.exists(paste0(query_url, ".html")))
```
jread-usgs, line 244: I like this, but I think it is the only use of `RCurl` in this repo/pipeline. Seems like a useful thing to include and I don't see the same type of function in another library we're already using (e.g., `httr`) so :+1: to have on.

jread-usgs (approved):

_targets.R
```diff
@@ -23,10 +23,10 @@ targets_list <- list(
...
     crs = "+proj=lcc +lat_0=45 + lon_0=-97 +lat_1=36 +lat_2=52 +x_0=0 +y_0=0 +ellps=WGS84 +units=m",
     cellsize = 25000,
-    xmin = -200000,
-    ymin = -1125000,
-    nx = 110,
-    ny = 85
+    xmin = -2700000,
+    ymin = -1750000,
+    nx = 217,
+    ny = 141
```
hcorson-dosch, line 29: Wow these are a big shift (25 grid cells in the y direction, and 100  in the x direction) -- are we no longer working solely with the midwest subset that the UW professor put together?

hcorson-dosch (approved):
This all looks great, Lindsay. I'm just curious as to why the grid parameters changed so much with the switch to the unbiased data -- seems like the footprint expanded?

jread-usgs:
Does the new grid correspond to the extent of this? 
image
(from [here](https://ccr.nelson.wisc.edu/dynamical-downscaling/index.php)) where the old version was a subset of it? Or is the new grid even bigger?

_targets.R
lindsayplatt: Not sure - these were the new parameters associated with the grid for `iy` and `jx`. I am wondering if the new projection is off? Because plotting this results in this:

```r
library(targets)
library(sf)

tar_load(grid_cells_sf)
grid_bbox_sf <- st_as_sfc(st_bbox(grid_cells_sf))

midwest_sf <- st_as_sf(maps::map('state', plot=FALSE, fill=TRUE)) %>% 
  st_transform(st_crs(grid_cells_sf))

plot(st_geometry(grid_bbox_sf))
plot(st_geometry(midwest_sf), add=TRUE)
```

image

lindsayplatt: @jread-usgs bigger than the extent you shared in that Temp figure. That figure looks like what the original extent we were using.

_targets.R
lindsayplatt: I am seeing max date is 2099-12-30 (but expect it to be 2099-12-31. I suppose I could write this as `2099-12-31 23:59:59` instead.

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -78,18 +78,10 @@ construct_grid <- function(cellsize, nx, ny, xmin, ymin, crs) {
...
+# TODO: Fancier version. This is a very rudimentary way to get all MN lakes.
 subset_lake_centroids <- function(centroids_sf) {
-
-  set.seed(19) # Same subset of 5 every time
-  wi_sf <- st_as_sf(maps::map('state', 'wisconsin', plot=FALSE, fill=TRUE))
-
   centroids_sf %>%
-    st_intersection(wi_sf) %>%
-    sample_n(5)
+    st_intersection(st_as_sf(maps::map('state', 'minnesota', plot=FALSE, fill=TRUE)))
```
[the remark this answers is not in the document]
lindsayplatt: I captured this "to do" as part of #271 but will need to prioritize it (at least for MN) soon. Do you see this as a polygon-to-polygon matching step to figure out which lakes are in which state using `canonical_lakes_sf` and then some spatial data for the states (like from Census Bureau)?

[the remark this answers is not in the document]
jread-usgs: Yes, I think census counties are the appropriate unit (which could be dissolved into states). In mntoha, this was: https://github.com/USGS-R/mntoha-data-release/blob/master/1_spatial.yml#L58

There is a question about whether you use the centroids (_within state_) or polygon (_overlaps part of state_)

In MNTOHA, the polygon overlap resulted in some lakes that belong to multiple states, since they overlap the borders (same with county ownership of a lake).

lindsayplatt:
^ @jread-usgs do you have any thoughts about the error I am seeing?

jread-usgs:
I think it is related to this: https://github.com/USGS-R/geoknife/issues/366

But shouldn't have any impact on the results, this is just an error with checking how far along the job is as far as I understand it. Looks like you still got data, which wouldn't happen if the job failed.

lindsayplatt:
Ah so I just need to update for now to not have it throw that error 👍

jread-usgs:
I'm assuming you already have that update in your pkg, but perhaps you are seeing another instance of that progress bar issue?

7_drivers_munge/src/GCM_driver_utils.R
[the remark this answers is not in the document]
lindsayplatt: Hmmm ok. I am trying to think of the implications for why we want to use this. We want to 1) be able to _subset data_ to see what we have for lakes _in_ certain states, and 2) want to be able to _summarize available data_ by state. If we decided to do polygon matching and allowed some lakes to belong to multiple states, we wouldn't need to remember that in downstream steps, e.g. if we wanted to model MN & WI we couldn't just filter by states to get the sites since there could be duplicates. We would need to also do a `unique()` before modeling. So would just need to note that next to the target.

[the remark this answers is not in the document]
hcorson-dosch: Not sure that we'd want to go this route, but we could have a table with a row for each lake and lake_id, then columns for each state, with a binary 0 or 1 for whether that lake is present in that state. Then for modeling multiple lakes, you could use AND or OR statements for filtering to get the subset of lake ids that you want. That way we wouldn't have duplicate rows.

lindsayplatt:
Oh, you are right. I did have the update. I ended up adding `gconfig('sleep.time' = 45, 'wait' = TRUE)` before my call and that got around the error. Should I use that or is it too temporary of a solution?

jread-usgs:
Yeah, but changing your sleep to 45 secs, you are going to reduce the number of times you ping GDP and thereby reduce the chance of you hitting this error. I assumed you were doing `wait` already, but just calling it within the `geoknife` function instead of using the config.

lindsayplatt:
I was running this:

```r
gcm_job <- geoknife(
  stencil = query_simplegeom,
  fabric = webdata(
    url = query_url,
    variables = c(
      'windspeed_debias', 
      'tas_debias', # air temp
      'rsds_debias', #shortwave radiation
      'rsdl_debias', # longwave radiation
      'rh_debias', # relative humidity
      'prcp_debias'), # precip
    times = c('1980-01-01', '2100-01-01')
  )
  # The default knife algorithm is appropriate
)
wait(gcm_job)
gcm_job_status <- check(gcm_job)$statusType
```

So doing the `wait()` right after? Sorry about confused about what the solution to getting passed this is?

jread-usgs:
basically `geoknife(..., wait = TRUE)` or a simple example
```r
job <- geoknife(stencil = c(-89,42), fabric = 'prism', wait = TRUE)
```
But setting the config value modifies whether wait is TRUE or FALSE by default.

7_drivers_munge/src/GCM_driver_utils.R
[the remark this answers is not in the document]
lindsayplatt: See #278

_targets.R
lindsayplatt: @wdwatkins pulling you in here to see if you know anything else about the grid differences between the original GCMs that were on GDP and the new debiased GCMs you added. I used the `iy` and `jx` variables to figure out the grid dimensions (see [here](https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1007807437)) but the new debiased GCMs have a much much larger grid scale than the previous grid (see [this fig](https://github.com/USGS-R/lake-temperature-model-prep/pull/276#issuecomment-1009149738) Jordan shared below). 

Do you have any understanding of why these might be so different? Is my projection off for the new grid? And if so, how might I find out the CRS from the GDP grid?

lindsayplatt: @hcorson-dosch and I tried to find the CRS from the new grid yesterday, but couldn't find one. Hayley explored the file using python and I tried with these R commands. Nothing is returned when it should be

```r
nc <- ncdf4::nc_open('http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias/files/debias_windspeed_mri_mid21_longerdaymet_ghcn_365d.nc')
ncdf4.helpers::nc.get.proj4.string(nc, "iy")
```

lindsayplatt:
Still waiting on my download to finish all the branches, but it has been marching along just fine so far. 30 out of 54 are complete. The only piece left is to verify the grid. I have added that as an item to #273, so that we can do that before calling our switch to the new GCMs complete. In order to get some of these grid and query (MN) changes to @hcorson-dosch to keep working on the NetCDF, we are going to merge this PR.

_targets.R
wdwatkins: Yeah, I don't see an explicit CRS anywhere in the debiased file I just checked.  You could sanity-check the extent by looking at the xlat/xlon variables though in the debiased files to take the projection out of the equation.  Just scrolling through the raw values your figure looks reasonable.

The GDP files have some metadata about grid origin etc in the global attributes.  You could also look at the raw lat/lon values there to confirm.
image


I would guess the GDP thredds data is subset to a particular study area if it is significantly smaller.

lindsayplatt: Hmmm but that is the old data set (the not debiased one). The new dataset doesn't show any Global Attributes http://gdp-netcdfdev.cr.usgs.gov:8080/thredds/dodsC/notaro_debias_mri.html

wdwatkins: Right.  You can always compare the xlat/xlon value between the two datasets though, and compare those to what the maps of projected extent look like .  One or both of those datasets may be subset from some larger file, to the extents may not be the same, but it would not be unreasonable that they use the same coordinate system.

lindsayplatt:
OK, the download finished!!! 🎉🎉🎉 Reminder that this download was for 9 tiles covering all of the MN lakes that we have in this pipeline so far. Each query was for all 3 of the projection time periods at once (21,900 days per query) and there were 6 GCMs (that's 54 queries in total!). Before sending to Hayley, I spent some time exploring. There are a couple of flags that I see:

1. The start/end times of the new debiased GCMs aren't the same as the old ones. Weirdly, the first chunk is shifted one year and all of them end on 12-30 rather than 12-31.
2. There are 4 cells that account for all of the NAs. Need to explore why these wouldn't have data.

### Map of the query

```r
# These libraries are needed for the rest of the code chunks
library(targets)
library(sf)
library(tidyverse)

tar_load(lake_cell_xwalk_df)
tar_load(grid_cells_sf)
tar_load(grid_tiles_sf)
tar_load(query_cells)
tar_load(query_tiles)

lakes_per_cell <- lake_cell_xwalk_df %>% group_by(cell_no) %>% summarize(nlakes = n())
mn_grid_tiles_sf <- grid_tiles_sf %>% filter(tile_no %in% query_tiles)
mn_grid_cells <- grid_cells_sf %>% filter(cell_no %in% query_cells) %>% left_join(lakes_per_cell)

ggplot() + 
  geom_sf(data = mn_grid_cells, aes(fill = nlakes)) +
  scico::scale_fill_scico(palette = "batlow", direction = -1) +
  geom_sf(data = mn_grid_tiles_sf, fill = NA, size = 2)
```

image


### Download diagnostics

Using the same function from [here](https://github.com/USGS-R/lake-temperature-model-prep/issues/258#issuecomment-999763969), I found out the following diagnostics for run time and file size. In total, it took 16 hours to download all the GCM data for all of MN. Seems OK, but not amazing. Not ideal as we scale, but I think we can move this part to HPC to help make it not as painful. Painful this time around only because I had an issue with the VPN connection timing out overnight & not letting my computer go to sleep. Running on HPC would fix both of those things. 

```r

# How long did the GDP download take and how big are the resulting files?
tar_load(gcm_data_raw_feather)
target_metadata_download <- summarize_target_build(names(gcm_data_raw_feather))

-- DURATION --
Shortest: 4.021 min
Longest: 44.123 min
Total: 988.789 min

-- FILE SIZES --
Smallest: 4.377 mb
Biggest: 95.938 mb
Total: 2116.269 mb
```

### Data diagnostics

```r
# Load and combine ALL downloaded GCMs (6 GCMs, 9 tiles w/ varying number of cells)
tar_load(glm_ready_gcm_data_feather)
munged_gcm_drivers <- purrr::map(glm_ready_gcm_data_feather, arrow::read_feather) %>% bind_rows()
```

#### Checking the data ranges (as described in item 1 above)

```r
# Range of each chunk:
munged_gcm_drivers %>% filter(time < as.Date('2030-01-01')) %>% pull(time) %>% range()
munged_gcm_drivers %>% filter(time > as.Date('2030-01-01'), 
                              time < as.Date('2070-01-01')) %>% pull(time) %>% range()
munged_gcm_drivers %>% filter(time > as.Date('2070-01-01')) %>% pull(time) %>% range()

"1981-01-01" "2000-12-30"
"2040-01-01" "2059-12-30"
"2080-01-01" "2099-12-30"
```

#### Understanding the NA values (as described in item 2 above)

```r
munged_gcm_nas <- munged_gcm_drivers %>% 
  mutate(across(.cols = matches("[A-Z]{1}[a-z]+", ignore.case = F), .fns = is.na)) %>% 
  rowid_to_column() %>% 
  filter_at(.vars = vars(-rowid, -time, -cell), .vars_predicate = any_vars(.))

# 4 cells account for all NAs when looking at data from all 6 GCMs
na_cells <- munged_gcm_nas %>% 
  group_by(cell) %>%
  summarize(count = n()) %>% 
  pull(cell)

# Where are these cells?
tar_load(grid_cells_sf)
tar_load(grid_tiles_sf)
tar_load(query_cells)
tar_load(query_tiles)

mn_grid_tiles_sf <- grid_tiles_sf %>% filter(tile_no %in% query_tiles)
mn_grid_cells_sf <- grid_cells_sf %>% filter(cell_no %in% query_cells)
grid_cells_na <- grid_cells_sf %>% filter(cell_no %in% na_cells)

ggplot() +
  geom_sf(data=mn_grid_cells_sf) + 
  geom_sf(data=grid_cells_na, fill = "orange") +
  geom_sf(data = mn_grid_tiles_sf, fill = NA, size = 2)
```

image

#### Checking the ranges of each variable

```r
daily_summary <- munged_gcm_drivers %>% 
  select(-cell) %>% 
  group_by(time) %>% 
  summarize_all(list(min, mean, max), na.rm = TRUE) %>% 
  pivot_longer(-time) %>% 
  tidyr::separate(name, c("variable", "fxn"), sep = "_") %>% 
  mutate(fxn = case_when(
    fxn == "fn1" ~ "min", 
    fxn == "fn2" ~ "mean", 
    fxn == "fn3" ~ "max"))
  
ggplot(daily_summary, aes(time, value, color = fxn)) +
  geom_line() + 
  facet_grid(variable ~ ., scales = "free_y")
```

Reminder of units: 

* Shortwave (W/m2)
* Longwave (W/m2)
* AirTemp (deg C)
* RelHum (%)
* WindSpeed (m/s)
* Rain (m/day)
* Snow (m/day)

image

lindsayplatt:
I have noted the remaining questions about this new GCM data (grid, dates, and missing cells) in the related issue #273. This code was approved prior to the download finishing (which it now has), so I am going to merge and allow those additional questions to be answered and potentially addressed in future PRs.

hcorson-dosch:
@lindsayplatt - at the moment I'm seeing GLM failures in lakes that cover [more than 4 cells](https://github.com/USGS-R/lake-temperature-process-models/issues/17#issuecomment-1020665443), but it might be helpful for my debugging purposes to know which 4 cells have NA values anyhow -- do you have that info?

lindsayplatt:
I am rerunning the code to get the cell numbers. Taking longer than I thought it would!

hcorson-dosch:
Update - the four cells are `c(17909, 18126, 18127, 18993)`. Thanks, Lindsay!

lindsayplatt:
Meant to tell you that I added those into the code above when I sent them earlier! Thanks for making sure they exist here, too!

PR #278, opened 2022-01-11
Add a lakes to state xwalk

lindsayplatt:
Added steps to `1_crosswalk_fetch` and `2_crosswalk_munge` to generate the target ``, which is a two-column table of `site_id` and `state` (the state abbreviation) indicating which state each site_id is in. Sites can be in multiple states, which would be represented by a separate row (so the site_id column represents non-unique site ids). I used a polygon-to-polygon intersection and any amount of intersection between a lake polygon and a state polygon. If the `state` column has an `NA`, this means it was not matched to any states (so likely Canada). There were 217 lakes the crossed multiple states and X lakes that were not matched to any state.

This fixes some of #271 and implements a need for #276 to use only MN sites. Tagging @padilla410 and @hcorson-dosch for awareness.

```r
library(scipiper)
library(tidyverse)

lake_centroids <- readRDS(sc_retrieve('2_crosswalk_munge/out/centroid_lakes_sf.rds.ind'))
xwalk <- readRDS(sc_retrieve('2_crosswalk_munge/out/lake_to_state_xwalk.rds.ind'))

head(xwalk)
# A tibble: 6 x 2
  site_id         state
  <chr>           <chr>
1 nhdhr_114337087 MN   
2 nhdhr_81804327  MT   
3 nhdhr_121544138 MN   
4 nhdhr_121544472 MN   
5 nhdhr_121545089 MN   
6 nhdhr_122546423 SD   

sum(duplicated(xwalk$site_id)) # 223 sites (0.27%) that crossed states
sum(is.na(xwalk$state)) # 10,918 sites (13%) that did not intersect any state boundaries

lake_centroids %>% 
  left_join(xwalk, by = "site_id") %>% 
  ggplot(aes(color = state)) + geom_sf()
```

image


1_crosswalk_fetch/src/fetch_zip_url_sf.R
```diff
@@ -0,0 +1,22 @@
...
+
+  destination = tempfile(pattern = layer_name, fileext='.zip')
+  file <- httr::GET(zip_url, httr::write_disk(destination, overwrite=T), httr::progress())
+  shp_path <- tempdir()
+  unzip(destination, exdir = shp_path)
+
+  us_counties_sf <- sf::st_read(shp_path, layer=layer_name) %>%
+    sf::st_transform(crs = 4326) %>%
+    mutate(state = dataRetrieval::stateCdLookup(STATEFP)) %>%
+    dplyr::select(state, county = NAME)
```
jread-usgs, line 15: nice. This will be useful

2_crosswalk_munge/src/spatial_crosswalks.R
```diff
@@ -188,3 +188,38 @@ buffer_sf_lakes <- function(out_ind, lake_ind, buffer_width){
...
+  # write, post, and promise the file is posted
+  data_file <- scipiper::as_data_file(ind_file)
+  saveRDS(conus_states, data_file)
+  gd_put(ind_file, data_file)
+}
+
+# Intersects lake polygons with state polygons to match each lake to
+# one or more states. If the lake doesn't match any (e.g. it's in
+# Canada), it is given an `NA` in the `state` column.
+crosswalk_lakes_intersect_states <- function(ind_file, lakes_ind, states_ind) {
```
jread-usgs, line 207: I'm just wondering if you want to skip having two targets, and put the `us_counties_to_states` call within this function. I know it would be normal to separate, but the actual meat of `us_counties_to_states` is pretty small and most of the code is the cache handling components, like the sc_retrieve/read, then save/write/put. 

What if instead you had this function accept `us_counties_ind` and had a TRUE/FALSE arg for summarize to county vs state? (I could see a future need for a county overlap target and you could use this function for both).

8_viz/src/plot_lake_obs_map.R
```diff
@@ -149,5 +130,5 @@ plot_lake_obs_map(
   fileout_base = '8_viz/out/lake_obs_map.png',
   temp_obs_ind = '7b_temp_merge/out/temp_data_with_sources.feather.ind',
   lakes_sf_fl = "1_crosswalk_fetch/out/canonical_lakes_sf.rds",
-  us_counties_sf = us_counties_sf,
+  us_states_ind = '2_crosswalk_munge/out/us_states_sf.rds.ind',
```
jread-usgs, line 133: 👍 good job bringing along this plot. I had forgotten this was in here.

jread-usgs (approved):

2_crosswalk_munge/src/spatial_crosswalks.R
lindsayplatt: Oh, I like that flexibility! I will work on editing

lindsayplatt: Updated! I also deleted the build yml, indicator, and GD data file related to that old target

PR #281, opened 2022-01-14
Add code to write munged GCM data to NetCDF files

hcorson-dosch:
This code writes the munged GCM data for all tiles (and all cells associated with those tiles) into a single NetCDF per GCM. The NetCDF files are written using [`write_timeseries_dsg`](https://github.com/USGS-R/ncdfgeom/blob/master/R/write_timeseries_dsg.R).

Marking this PR as draft for three reasons:
1) @lindsayplatt needs to test this to confirm it runs with the actual targets, as I had to use a placeholder target for the `glm_ready_gcm_data_feather` files and provide those files manually
2) We're not 100% certain that the current netCDF format conforms to NetCDF CF conventions - see discussion in #252 
3) We are still investigating some issues with the grid projection, dates, and NA values -- see #273


hcorson-dosch:
Current plan is to go ahead and get this reviewed and merged into the `gcm_driver_data_munge_pipeline` branch of of the canonical repo, despite the unresolved issues noted in #273, so that I can get a start on pulling the data from the resulting netCDF files for use in [`lake-temperature-process-models`](https://github.com/USGS-R/lake-temperature-process-models/blob/main/1_prep.R#L24-L39).

@jread-usgs -- as discussed I'm leaving in [my work-around](https://github.com/hcorson-dosch/lake-temperature-model-prep/blob/gcm_driver_data_munge_pipeline/_targets.R#L175-L181) that uses manually provided feather files, rather than requiring the munged feather files generated using the raw downloaded feather files, which are only on Lindsay's local machine. You'll need to unzip [this archive](https://doimspp.sharepoint.com/:u:/r/sites/IIDDStaff/Shared%20Documents/Project%20-%20Modeling%20Lakes/GLM/lake-temperature-process-models/glm_ready_gcm_data_feathers.zip?csf=1&web=1&e=Jkbfb6) into `'7_drivers_munge/tmp'`, which should then allow you to build the `glm_ready_gcm_data_feather` targets that you'll need to build the `gcm_nc` target.

hcorson-dosch:
Also you'll see I edited the map_query code to produce the map used by @lindsayplatt [here](https://github.com/USGS-R/lake-temperature-model-prep/pull/276#issuecomment-1013264491), since it seemed like a useful one.

jread-usgs:
Cool. I'm able to run, but ran into this error 

```r
• built branch query_cells_centroids_list_by_tile_93a603de
• built pattern query_cells_centroids_list_by_tile
• start target query_map_png
Joining, by = "cell_no"
x error target query_map_png
• end pipeline
Error : object 'grid_tiles_sf' not found
```
When starting to build the .png. Is there a place where that target needs to be defined?

hcorson-dosch:
Oh just a typo - didn't catch it b/c it was in my environment. Sorry. Will commit the fix now

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -3,25 +3,99 @@
...
+    # https://github.com/USGS-R/ncdfgeom/blob/master/R/write_timeseries_dsg.R
+    ncdfgeom::write_timeseries_dsg(nc_file,
+                         instance_names = gcm_cells,
+                         lats = cell_centroid_lats,
+                         lons = cell_centroid_lons,
+                         alts = NA,
+                         times = gcm_dates,
+                         data = var_data,
+                         data_unit = var_data_unit,
+                         data_prec = var_data_prec,
```
jread-usgs, line 89: I see this is coming in as `"float"`. Is that more appropriate than `"double"` for these data?

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -3,25 +3,99 @@
...
+    var_data_metadata <- list(name = variable_metadata$var_name, long_name = variable_metadata$longname)
+
+    # Check if nc file already exists, and therefore should be added to,
+    # or if it needs to be created
+    var_add_to_existing <- ifelse(file.exists(nc_file), TRUE, FALSE)
+
+    # write this variable to the netcdf file:
+    # ncdfgeom::write_timeseries_dsg() function documentation:
+    # https://github.com/USGS-R/ncdfgeom/blob/master/R/write_timeseries_dsg.R
+    ncdfgeom::write_timeseries_dsg(nc_file,
```
jread-usgs, line 81: One think to think about when arranging the data is how it is likely to be accessed in chunks, or contiguous parts of the `gcm_cells` instances. It should be some kind of logical arrangement, like left to right, bottom to top. I see you've done that here:

image
(redder are instances closer to the beginning of the vector)
This looks good to me, but putting this check in so you can see it. I used my old-school
```r
plot(cell_centroid_lons, cell_centroid_lats, col = colorRampPalette(c("red", "blue"))(length(cell_centroid_lats)))
```

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -3,25 +3,99 @@
...
+                         data_prec = var_data_prec,
+                         data_metadata = var_data_metadata,
+                         time_units = data_time_units,
+                         instance_dim_name = gcm_cells_dim_name,
+                         dsg_timeseries_id = gcm_cells_dim_name,
+                         attributes = data_attributes,
+                         coordvar_long_names = data_coordvar_long_names,
+                         add_to_existing = var_add_to_existing,
+                         overwrite = overwrite)
+  }
```
jread-usgs, line 98: One thing to consider is whether you'd benefit from compression of this file, or if it is necessary ([see system calls here](https://github.com/jread-usgs/lakesurf-data-release/blob/main/src/ncdf_utils.R#L85))

jread-usgs (commented):
Some minor comments. Strongest suggestion is around the file and dim naming.

_targets.R
```diff
@@ -90,28 +93,30 @@ targets_list <- list(
...
   # TODO: maybe remove this once we are happy with this process
-
-  # Save image of each map query for exploratory purposes
+  # Save image of full map query for exploratory purposes
   tar_target(
     query_map_png,
     map_query(
-      out_file_template = '7_drivers_munge/out/query_map_tile%s.png',
-      lake_centroids = query_lake_centroids_sf,
+      out_file = '7_drivers_munge/out/query_map.png',
```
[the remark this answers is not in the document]
hcorson-dosch: Yeah, agreed. Kudos to Lindsay for creating it!

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -3,25 +3,99 @@
...
+
+  # Get vectors of cell centroid latitudes and longitudes
+  cell_centroid_lats <- cell_coords %>% pull(lat)
+  cell_centroid_lons <- cell_coords %>% pull(lon)
+
+  # Set up attributes for NetCDF that are independent of variables
+  data_time_units <- "days since 1970-01-01 00:00:00"
+  data_attributes <- list('title' = global_att)
+  data_attributes <- append(data_attributes, grid_params) # add grid parameters to list of global attributes
+  data_coordvar_long_names <- list(instance = "identifier for GCM grid cell", time = "date",
```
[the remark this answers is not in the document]
hcorson-dosch: oooh good point.

7_drivers_munge/src/GCM_driver_nc_utils.R
hcorson-dosch: Hmm I can make the switch to `"double"` - I see that's what you and Jared used [here](https://github.com/jread-usgs/lakesurf-data-release/blob/main/src/ncdf_utils.R#L148)

7_drivers_munge/src/GCM_driver_nc_utils.R
hcorson-dosch: Yes, good point. I'd seen that in your original code, and figured we'd revisit once this code was up and running. Lindsay and I checked and right now the raw feather files total 1.91 gb and the netCDF fiels total 1.24 gb, so it's a reduction but not a huge reduction. How much more of a reduction did compression get you in the `lakesurf-data-release`?

jread-usgs: Hmmm...I checked on this one with this
```
ncks -h --fl_fmt=netcdf4 --cnk_plc=g3d --cnk_dmn time,10 --ppc Rain=.3#Snow=.3#AirTemp=.2#RelHum=.1#Shortwave=.1#Longwave=.1#WindSpeed=.3 7_GCM_MRI.nc compressed_MRI_out.nc
```
And `7_GCM_MRI.nc` (original file) is 223.4 MB and `compressed_MRI_out.nc` is 72 MB for me. The ".3" and ".1" are the different precisions I set for the different variables as a starting point. Otherwise I was using the existing code from that other repo.

7_drivers_munge/src/GCM_driver_nc_utils.R
hcorson-dosch: ooh this is impacting file size (and build time) quite a bit. Now the netCDFs are 436 mb each, before they were 213 mb each.

hcorson-dosch: This build is _still going_. Taking at least 5x as long as before.

jread-usgs: I think I'm going to backtrack on my comment :) since double probably has an unnecessary level of precision, and if we're further reducing precision at the end of the .nc build (see `nco` comment below), it isn't going to help to use `double` here. `float` seems better all around 👍

jread-usgs:
One comment I wanted to add here that could be discussed in the future is what form the final data take with respect to the different GCMs that are debiased/downscaled. 

Currently, this PR has a separate file for each GCM, which is consistent with what we've done in the past (e.g., GCM-specific zip files in the Winslow release). But one advantage of NetCDF is that it would allow us to use additional dimensions that could cover things like different GCMs...So an alternative would be to have one .nc file that contains all the data, with a dimension for each GCM. This would make it easier to get data from one variable from all GCMs at the same time. 

I'm asking this because I think the files aren't going to be brutally huge and often users of this (like us!) will want to access _all_ downscaled data for a given cell vs go to different files for that. Perhaps something to discuss later down the road when we're making decisions about the data release structure.

7_drivers_munge/src/GCM_driver_nc_utils.R
hcorson-dosch: @jread-usgs - I've added in the code to write to temporary files and then compress those, but can't test it on my machine because I don't have the command-line programs to do so. Is [NCO](https://github.com/nco/nco) what I need?

jread-usgs: Yes, nco and the netcdf libraries should do it, but I bet you already have the netcdf libraries installed. I use homebrew to install these on mac, but not sure what to recommend for windows.

hcorson-dosch: Hmm I installed via conda-forge, which seems to be what is recommended for Windows [here](http://nco.sourceforge.net/). I can now compress the files with the ncks command in an Anaconda Prompt shell, but not from R, within the pipeline. 🙁.

For me the files go down from 213mb to 68.7mb

jread-usgs: @dblodgett-usgs got install recommendations for Hayley in this case?

hcorson-dosch: Per [the README.md of this R package wrapper of NCO](https://github.com/markpayneatwork/ClimateOperators), which I've not yet attempted to install, just stumbled upon, I've been trying to make the conda installation of NCO [available in my system path](https://github.com/markpayneatwork/ClimateOperators/blob/master/README.md?plain=1#L15), so that I can call it from R using system commands. I'm a bit out of my depth here, but when I added `C:\Users\hcorson-dosch\Anaconda3\pkgs\nco-4.8.1-habc9422_100\Library\bin` to my PATH environment variable, I now no longer get a `ncks: command not found` error in Git Bash, but I do get this error: 
```
hcorson-dosch@IGSARMEWLTHCOR MINGW64 ~/Documents/R/lake-temperature-model-prep/7_drivers_munge/out (gcm_driver_data_munge_pipeline)
$ ncks
C:/Users/hcorson-dosch/Anaconda3/pkgs/nco-4.8.1-habc9422_100/Library/bin/ncks.exe: error while loading shared libraries: api-ms-win-crt-environment-l1-1-0.dll: cannot open shared object file: No such file or directory
```

hcorson-dosch: Okay hit a whole series of missing .dll files, most of which I was able to copy to the directory of the `ncks.exe` from `C:\Users\hcorson-dosch\Anaconda3\Library\bin`, but not having luck finding `udunits2.dll` which it now can't find.

If I now attempt to run the pipeline and compressed the files through RStudio, I get these error messages:
image
image

hcorson-dosch: Alright I've installed `udunits2` from CRAN, and copied `udunits2.dll` to `C:\Users\hcorson-dosch\Anaconda3\pkgs\nco-4.8.1-habc9422_100\Library\bin` from `C:\Users\hcorson-dosch\Anaconda3\pkgs\udunits2-2.2.27.27-h892ecd3_3\Scripts`, which has fixed that first error.

lindsayplatt: Not a comment to help out with this error (because I have no idea), but let's make sure we note these external libraries somewhere and maybe log an issue to add a helpful error message if you are missing them in the future.

lindsayplatt: Another thought: while compression is important in the end and a general good idea/best practice, let's also keep in mind practicality for our timeline. We want to get some results up by the end of the week and maybe we can't get a compressed NetCDF by then. In order to keep moving forward, maybe we log an issue for adding file compression, link to this thread, and then keep moving forward with an MVP?

hcorson-dosch: Alright just added another commit that makes `compression` a T/F argument to the `generate_gcm_nc()` function. If it is set to TRUE and fails, the function will throw an error. Note: I tested but didn't use a tryCatch because if NCO isn't set up R throws a system error, but not a console error, so it isn't caught by TryCatch. So instead I'm using a simple check to see if the output compressed file was created, and, if not, throwing an error.

hcorson-dosch: @jread-usgs - can you confirm that the compression code works on your end? Note you'll need to change the `compression` argument in the target recipe for `gcm_nc`, to `TRUE`.

If it does work, I'll add another commit deleting my workaround for the `glm_ready_gcm_data_feather` files, and then @lindsayplatt, could you test the code on your end (with `compression` = `FALSE`) to make sure the pipeline builds using the actual pipeline built `glm_ready_gcm_data_feather` files, rather than those I manually provided?

jread-usgs (approved):
I'd suggest filing an issue to come back to the compression part of this and do two things: 1) break out the compression code into another function that does the compression and nothing else, and 2) add information on the system libraries needed and how to get them. 

For the time being I see no issue keeping compression off and would even go so far as to say you could put a `stop()` right at the top of the function for when compression == TRUE, until it is supported deeper. I have done that in the past where I haven't fully covered the cases in the arguments, but still want to work with a version of the function that covers our MVP. 

Yes, can confirm that when I set the compression flag to `TRUE` I was able to build these files in their compressed forms (they ended up 71-72.5MB compressed and 223.4MB w/ compression == FALSE)

hcorson-dosch:
Okay added a `stop()` argument if `compression == TRUE` to note that the option is not yet fully supported

hcorson-dosch:
Okay @lindsayplatt can you just confirm that this builds as is for you with the actual `glm_ready_gcm_data_feather` files associated with that original target?

7_drivers_munge/src/GCM_driver_nc_utils.R
dblodgett-usgs: For NetCDF and nco, I think you can install using chocolaty. I would say run with Conda if it works, but have never had good luck using Conda-installed-dependencies outside of a Conda environment. 

I thought I'd installed it with chocolatey -- but now I can't find that source. 

I think you could probably just install the precompiled binary NetCDF (netCDF4+DAP) and use nccopy. nccopy does chunking and deflation. 

`nco` is not available via precompiled binary AFAIK, and it required for that --ppc lossy compression scheme. I use WSL linux or a dockerized (linux) NetCDF to do that kind of stuff rather than messing with getting it compiled and on my path in Windows.

jread-usgs: I didn't realize this was such a headache on Windows...

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -3,25 +3,141 @@
...
+                         dsg_timeseries_id = gcm_cells_dim_name,
+                         attributes = data_attributes,
+                         coordvar_long_names = data_coordvar_long_names,
+                         add_to_existing = var_add_to_existing,
+                         overwrite = TRUE)
+  }
+
+  # If compressing netCDFs, compress the temporary netCDF file and save
+  # to the final output file
+  if (compression == TRUE) {
```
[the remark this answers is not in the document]
hcorson-dosch: FYI I created that issue yesterday - #294

7_drivers_munge/src/GCM_driver_nc_utils.R
dblodgett-usgs: Theoretically, the conda-forge solution should work... I have just never taken the time to figure out how to get it working correctly.

hcorson-dosch:
Thanks, Lindsay. That missing value is [hardcoded](https://github.com/USGS-R/ncdfgeom/blob/main/R/write_timeseries_dsg.R#L118-L120) in the `write_timeseries_dsg()` function.

And that is odd about the ranges of those variables -- I can do a check of the munged feathers.

lindsayplatt:
Huh that is just a really interesting way to denote missing values. Nevermind then :) If the ranges in the feathers are the same, I'll log that issue to #273 and we can move this one along.

hcorson-dosch:
Okay working through this with the munged feather files.

I do see NA values for all variables, but it looks like maybe some of the data variables use NA to denote NA values, while some use NaN? I wonder if that's not picked up in the conversion to netCDF? For example, if I leave out `na.rm=TRUE` in my calls within `summarize()`, which I did by accident the first time round, I see this:

image

Lindsay is digging into this now.

The number of NA values for each variable is 52560. For each variable, for each GCM, there are 87600 NA values. 87600 days / 365 days per year = 240 years of NA values. We know we have 60 years of data in total, so that means those 240 years of NA values could be attributed to the 4 cells with NA values. If I exclude those four cells `c(17909, 18126, 18127, 18993)` then I see 0 NA values for all variables, for all GCMs, so that tracks.

If I summarize the results across all gcms, and exclude NA values, I get this:
image

Here's the whole tibble, broken down by GCM
image

lindsayplatt:
Comparing the munged feathers (Hayleys values above) for just GFDL with those from Winslow 2017 confirms that all the values in the munged feather files are within reason.

WINSLOW GFDL:
image

This pipeline GFDL:
image

Now to just figure out why the values in the NetCDF are different.

dblodgett-usgs:
I you are using ppc compression for this, some change to the values is expected. I lost track of what all transformations you are making.

hcorson-dosch:
Alright I think Lindsay and I got to the bottom of these discrepancies. We confirmed that in the munged feathers, all variables *except* Rain and Snow coded NA values with `NaN`, but Rain and Snow coded NA values with `NA`
image

Once that data was written to a netCDF and then pulled, we were seeing weird max values for Rain and Snow, as Lindsay noted above:
image

If I replace all `NA` values with `NaN` _before_ writing to netCDF, then when I use Lindsay's code to pull those same values back out, the ranges are reasonable:
image

And match the summary derived from the munged feathers:
image

So looks like `RNetCDF` expects `NaN` values

PR #282, opened 2022-01-18
GCM munge pipeline feathers split by time (part of temp workaround to NetCDF-DSG)

lindsayplatt:
This is a change to split the resulting feather files (intermediate targets that will go away once we figure out #281) to be split by cell and time period, so that they can more readily be used in the GLM pipeline.


hcorson-dosch (approved):
Thanks, Lindsay!

lindsayplatt:
Finally committed my last local change to get this to work. Merging since Hayley already reviewed and this last change was a small switch to branching

PR #284, opened 2022-01-21
Add iadnr data

padilla410:
Data summary:
* 176 unique lakes in the raw data and  113 in `all_coop_dat_linked.feather` (64% of the lakes in the raw data)
* 603,170 records in the raw data and 491,109 in `all_coop_dat_linked.feather` (~81% of the records in the raw data)

I think the major reason for the loss here is the mismatch between the raw data and the existing lookup table.
The existing look up table has 132 lakes while the raw data has 176. We could easily pick this up if we had the rest
of the spatial info.

A few checks:
```r
library(scipiper)
library(arrow)

# load current version of data set on gdrive
dat_gdrive <- read_feather(sc_retrieve("7a_temp_coop_munge/out/all_coop_dat_linked.feather.ind"))

# filter for recent IA DNR data
dat_iadnr <- dat_gdrive %>% 
  filter(source == '7a_temp_coop_munge/tmp/Iowa_DNR_LimnoProfiles_2000_2020.rds')

# check record count
nrow(dat_iadnr)

# check lake count
length(unique(dat_iadnr$state_id))
```

the outputs:
```r
> nrow(dat_iadnr)
[1] 491109

> length(unique(dat_iadnr$state_id))
[1] 113
```

Snapshot of `8_viz/out/lakes_summary_fig.html`
image


7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+  data_clean <- dplyr::bind_rows(dat_2000_2003,
+                          dat_2004_2007,
+                          dat_2009,
+                          dat_2010,
+                          dat_2011_2016,
+                          dat_2017b,
+                          dat_2018_2020_1,
+                          dat_2018_2020_2) %>%
+    dplyr::mutate(Iowa_ID = paste('Iowa_', lakeid, sep = ''), # for matching the xwalk
+      DateTime = as.Date(date)) %>% # there is a mix of POSIXct and Date in the indiv dfs
```
jread-usgs, line 147: Interesting...if `date` has different classes, I'd assume the `bind_rows` would either fail or coerce them all into the same. E.g.,
```r
bind_rows(tibble(data = 2), tibble(data = "4"))
Error: Can't combine `..1$data` <double> and `..2$data` <character>.
```
Similar to my comment above on temperature as °C or °F, it seems each of the individual parsers would be the place to ensure a common unit/class for what is coming out of it, so that the bind is simple (I do, however, agree that the `Iowa_ID` should happen here once, which makes it more future-proof if we for some name modify that ID name in the future, we'd only have to change it here for the parsers).

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+    dplyr::select(! c(lake, lakeid, date)) #holdover column from 2017-2020 data
```
jread-usgs, line 148: If it isn't too wordy, would you consider `select` where you are naming the columns you are keeping instead of the ones you are removing? That is a bit easier to read and understand what the final output is w/o digging into the code. If the columns are too numerous though, it doesn't make sense.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+
+  dat <- dat %>%
+    dplyr::select(all_of(tolower(keep_cols))) %>%
+    dplyr::mutate(depth = as.numeric(depth), temp = as.numeric(temp))
+
+  # add LakeID and date columns from file name if necessary
+  dat$lakeid <- extract_lakeid(x, start = subset_lakeid[1],
+                               stop = subset_lakeid[2])
+
+  # wrangle date and time
```
jread-usgs, line 191: Yikes there is a lot of variety in here!

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+    # check max depth and trim data if necessary
+    if(check_final_depth(dat$depth)) {
+      depth_loc <- which(names(dat) == 'depth') # id position of the depth column
+      dat <- trim_to_max_depth(dat, depth_col = depth_loc)
+
+      # return(dat)
+    }
+  } else {
+    print(paste('File ', basename(x),
+                ' appears to be empty or has problematic data. Data only has ',
```
jread-usgs, line 225: I had 
```r
[1] "File 2005_L113_2005173014.xls appears to be empty or has problematic data. Data only has 1 row(s) of data" 
```
I prefer `message` to `print` in these situations, perhaps switch to that?

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+                dat_rows, ' row(s) of data', sep = ''))
+  }
+
+  return(dat)
+}
+
+#' Determine if a file is `xls` or `htm`
```
jread-usgs, line 232: 😱

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+  }
+
+  # check for datetimes and AM/PM designation
+  if(any(sapply(dat$time, function(x) class(x) == "POSIXct"))) {
+
+    dat$time <- strftime(dat$time, format = "%H:%M") %>% as.character()
+
+  } else if(any(sapply(dat$time, function(x) nchar(x) > 9))) {
+
+    dat$time <- ifelse(nchar(dat$time) == 11,
```
jread-usgs, line 317: Perhaps add a comment above this to tell us what the `ifelse` is looking for

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+  dat_2017b <- files_2017b %>%
+    purrr::map_dfr(~ parse_2017_2020_data(.,
+                                          keep_cols = c('DATE', 'TIME', 'SITE',
+                                                        'Temp...F.', 'Depth..m.'),
+                                          lakeid_col = 3,
+                                          temp_as_f = T,
+                                          use_readr = F)) %>%
+    dplyr::mutate(Timezone = NA) %>%
+    dplyr::rename(depth = depth..m.,
+           temp = temp...f.) %>%
```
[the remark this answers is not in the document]
jread-usgs: Nevermind, I see below what `temp_as_f` is doing. Didn't catch that initially. Looks good.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+#' @param x chr, full file path
+#' @param keep_cols chr, vector of column names to keep.
+#' @param lakeid_col num, Column index for the location of the Lake ID field. The column that is used to ID the Iowa lake moves around and changes names from file to file.
+#' @param temp_as_f logical, should temp values be converted from F to C?
+#' @param use_readr logical, not all files are read well by `readr::read_csv`. If `use_readr = F` then `read.csv` is used.
+#' Column name spelling should match. Case does not matter because
+#' all values to be changed `tolower` within the function.
+parse_2017_2020_data <- function(x, keep_cols, lakeid_col, temp_as_f = F, use_readr = T){
+
+  if(grepl('*xlsx', x)) {
```
jread-usgs, line 278: Since `x` is used so much in this function, perhaps use a more intuitive variable name, like `filepath`?

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+    dat <- readxl::read_xlsx(x)
+  } else if (use_readr){
+    dat <- readr::read_csv(x)
+  } else {
+    dat <- read.csv(x)
+  }
+
+  # select key columns, normalize
+  dat <- dat %>% dplyr::select(all_of(keep_cols))
+  names(dat) <- tolower(names(dat))
```
jread-usgs, line 288: I'm surprised that you don't need to normalize/`tolower()` before you remove columns based on names if there is variety in the naming.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+    # 2018-2020 data has a little extra whitespace and often includes the seconds
+    dat$time <- format(dat$time, '%H:%M') %>%
+      as.character(.) %>%
+      substr(., 1, 5)
+
+  }
+
+  # convert temp if necessary
+  if(temp_as_f) {
+    idx <- grepl('*temp*', names(dat)) %>% which(. %in% "TRUE")
```
jread-usgs, line 332: As I'm digging into these files, I see why all of the funky partial matching is necessary 😱

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
jread-usgs: Yeah, below in line [365](https://github.com/USGS-R/lake-temperature-model-prep/pull/284/files#diff-c1aa506bc19f1074218a6700aecfa8c979ff5a24f33dd1c4cd2f2791b286b979R365) you use `dplyr::select(all_of(tolower(keep_cols)))` (adding `tolower()` there). Perhaps you need the same here, but `tolower()` the names first?

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+#' Trim the lake profile to the maximum depth
+#'
+#' The `explain` file for 2000-2003 Iowa DNR data (`Iowa_profiles_2000-2003.docx`)
+#' notes that some datasets contain data from the descent _and_ the ascent of the data
+#' logger. The ascent data should be trimmed.
+#'
+#' @param df data.frame that includes the depth
+#' @param depth_col int, index for depth column. Defaults to the first column (`depth_col = 1`).
+#'
+trim_to_max_depth <- function(df, depth_col = 1) {
```
jread-usgs, line 456: I don't think we should do this within the parsers. I think separately we want to compare reported depth against profile depths, but I don't think we should screen within the functions. Sometimes the reported depths are wrong and sometimes the water levels fluctuate.

jread-usgs (commented):
whooo, that was an adventure 😓 
Some questions and comments inline here

A general question, for the partial matching and indexing and reassigning with base R, is there a way to do some of these operations with `case_when` and tidyselect matching like `contains` and `starts_with` to keep things dplyr'd?

jread-usgs:
I end up with an error when trying to parse the files
```r
Error in make.names(col.names, unique = TRUE): invalid multibyte string 5; debug with scmake("7a_temp_coop_munge/tmp/Iowa_DNR_LimnoProfiles_2000_2020.rds.ind", "7a_temp_co...
```
I will try to track down which file that is. Perhaps a windows/mac difference on one of the parsers/files?

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: Absolutely

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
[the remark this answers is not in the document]
padilla410: Yes, the 2017 data is the only set in fahrenheit. I debated renaming the column but thought that would be too much of an unexpected surprise. However, I _will_ update the doc to note that the col name doesn't change.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: I *think* it works because the two classes are `Date` and `POSIXct` and they're close enough (i.e., `dplyr::bind_rows` can handle two similar dates by coercing one to another). I did it at the end as a kludge because the pipeline was breaking when binding the dates. 

As you noted, the most correct thing to do would be to clean it up within the individual parsers, which I will.

jread-usgs:
`Iowa_DNR_LimnoProfiles_2000-2020/Iowa_profiles_2017b/17129263ysi.csv` causes the error for me

it is addressed by adding the fileEncoding to `read.csv`
```r
read.csv(x, fileEncoding="latin1")
```
in the `else` of `parse_2017_2020_data()`

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+#' Column name spelling should match. Case does not matter because
+#' all values to be changed `tolower` within the function.
+parse_2017_2020_data <- function(x, keep_cols, lakeid_col, temp_as_f = F, use_readr = T){
+
+  if(grepl('*xlsx', x)) {
+    dat <- readxl::read_xlsx(x)
+  } else if (use_readr){
+    dat <- readr::read_csv(x)
+  } else {
+    dat <- read.csv(x)
```
jread-usgs, line 283: Would you be up for changing this `read.csv` to `read.csv(x, fileEncoding="latin1")`? Or does that fail on your windows machine for "Iowa_DNR_LimnoProfiles_2000-2020/Iowa_profiles_2017b/17129263ysi.csv"?

jread-usgs: See https://github.com/padilla410/lake-temperature-model-prep/pull/1

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
```diff
@@ -0,0 +1,507 @@
...
+    dplyr::mutate(Timezone = NA) %>%
+    dplyr::rename(depth = depth..m.,
+           temp = temp...f.) %>%
+    dplyr::select(date, time, Timezone, depth, temp, lakeid)
+
+  # Clean data from 2018-2020 ---------------------------------------------------
+  ## `2019_R3_Profiles.csv` is in an unexpected format. it was faster to just munge
+  ## separately
+
+  dat_2018_2020_1 <- files_2018_2020[c(1:5, 7:9)] %>%
```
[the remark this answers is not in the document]
padilla410: Good point. I will address both of these points.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: In a brief side bar conversation, Jordan and I decided to keep the `trim_to_max_depth` for two reasons:
1. Only a few of the years have this problem, and 
2. IADNR mentions that we should do this in several docs because data includes values from the ascent of the probe (e.g., [2000-2003](https://docs.google.com/document/d/1hhBk6Oz35XRR03jfRvBCXI9XTkLd3yYV/edit) and [2004-2007](https://docs.google.com/document/d/1nr_ODU9IQ_7ns6o960MLekRx84M-UBbi/edit#heading=h.gjdgxs)). These values did not equilibrate.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: Should I keep it or shift to `case_when`? In general, I try to use `base` solutions if aren't too convoluted

padilla410: Also, I'll add some more doc here

jread-usgs: Complicated question ("Should I keep it or shift")! 

I generally try to stay in dplyr-land for these munging codes since it is easier to read and understand in general. But I do also agree that we should stay in base when there is an adequate solution and no good reason to bring in a _new_ package to provide a single function. I think our general (unwritten ☹️ ) coding styles for the branch would suggest this munging code would be in dplyr, plots are generally ggplot2, but all kinds of space to deviate from that pattern as needed. 

But this is probably something others should weigh in on. At the project lead level, @lindsayplatt do you have opinions? I also think there is room to suggest a future pattern without asking for a re-write of this particular code. This is related to my [overarching comment above](https://github.com/USGS-R/lake-temperature-model-prep/pull/284#pullrequestreview-859450328) that was
> A general question, for the partial matching and indexing and reassigning with base R, is there a way to do some of these operations with case_when and tidyselect matching like contains and starts_with to keep things dplyr'd?

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: I resolved this issue throughout all of the functions

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: Got it. I added it to the function and tested it on my end. I am going to reject your PR into my repo because I will resolve it here.

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: Resolved

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: :D A wild ride, but clean data feels good, man

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
lindsayplatt: I have been staying in tidyverse-land lately with projects due to its readability, but I think its true that there is a critical point when you are starting a project where you need to think about the benefits of loading in a new package(s). For this project, we already use `dplyr` and other `tidyverse` functions in a number of places, so I think it makes sense to lean into that practice. Times when we might want to deviate are when the dplyr commands are significantly slower (which can happen with some bigger data). In those cases, we may turn to base R or to something like `data.table`.

padilla410: I have definitely been working to avoid new packages. How about we split the difference on this one and I start using `case_when` into the future. I suggest that for this specific case because there are so many edge cases in this dataset (hence all the partial matching and `ifelses`) within this dataset. It would take a bit of time to resolve it. To compensate for any clarity issues I can add more into the comments.

lindsayplatt:
I just did a before/after with the new HTML and the old one. Iowa is on the map for PGDL now 🎉

image

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
lindsayplatt: I think extra comments in lieu of converting to `case_when()` would be a good compromise in this instance. Maybe also make a note about this very decision somewhere in the code - choosing to forgo a conversion to `case_when()` in order to keep moving forward (could even link to this discussion by clicking on the time next to the username to get a link specifically to these comments).

jread-usgs: I'm good with this too and hope my original comment was worded such that it was clear I thought this change would be totally optional here, but something to consider in the future. Glad we had this discussion!

padilla410: Sidebar chat with Lindsay. We are going to leave this as is within this parser, but I will use `case_when` in future parsers

7a_temp_coop_munge/src/data_parsers/parse_Iowa_DNR_LimnoProfiles_2000_2020.R
padilla410: Yes, you're right. I just got lucky on these files.

padilla410:
OK! I resolved all of the comments and successfully ran `scmake(7a_temp_coop_munge)` locally (`options(scipiper.dry_put = TRUE)`). I intend to do a run that talks to gdrive after a finish a few summary tasks for a 3 pm ET meeting today.

jread-usgs (approved):
I've been following the changes/commits as they came in and I approve this PR. 🎉

PR #292, opened 2022-01-27
adding lingering build files

jread-usgs:


jread-usgs:
I reviewed this, we discussed it

PR #297, opened 2022-02-04
Update GCM grid using jx/iy as centroid not bottomleft

lindsayplatt:
Through a series of comments/investigations (starting with [this comment](https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1028474585)), we discovered that we were setting up the grid to query GDP incorrectly. It was slightly shifted because we had made the assumption that the grid points provided (described as "cross points" in the attribute information) were the _bottom left corner_ of the grid cells when they were indeed the _centroid_. This PR edits the grid construction functions to shift the points 0.5 of the cell width & height in order to get the correct x/y for the bottomleft corner of the bottomleft most grid cell.

Related to fixing #273 

Here was the `query_map_png` before this fix (taken from [here](https://github.com/USGS-R/lake-temperature-model-prep/pull/276#issuecomment-1013264491))

image

Here is the `query_map_png` target after this fix. Still 9 tiles 

query_map


lindsayplatt:
Separately running a quick test to compare download times if I did a 15x15 tile instead of 10x10 ([per Jordan's comment here](https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1020197209)) since I have to re-download everything with this new grid configuration anyways. That potential change will be in a different PR.

hcorson-dosch (approved):
This looks good. I still think it's so odd that the netCDF attribute info refers to the `xlat` and `xlon` as "Latitude on Cross Points" and "Longitude on Cross Points" if they are actually referring to centroids....

Issue #279, opened 2022-01-14
Scale GCM munge pipeline to whole footprint


lindsayplatt:
Learning from the experience noted in #258 and documented diagnostics after building the pipeline for all of MN here, this task will need to do two things:

1. Move the pipeline processing onto HPC resources. Likely Denali so that the resulting data can easily be leveraged by our [lake-temperature-process-models](https://github.com/USGS-R/lake-temperature-process-models) code which is also running on Denali.
2. Change the `query_lake_centroids_sf` target to subset to the appropriate footprint (MGLP or everything that we have in here?)
3. Change the tile size? All of MN could probably be 1 tile. See @jread-usgs's comment here https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1020197209

It took 16 hours to build all of MN. There could be additional opportunities for optimizing the downloads. Exploring that could be part of this issue, or we could accept the lengthier builds and just kick it off on HPC resources, knowing that it will take a few days.

lindsayplatt:
So far, I have 

-  Moved the munge over to Tallgrass (using an interactive singularity container adapted from `lake-temperature-out` right now, but hoping to adjust soon)
-  Adjusted the lakes we are using to be based on lake ids in the NML list
-  Tested and chose a new tile size (see below)
-  Looked into fixing #280, but did not change anything because the newest version isn't out on CRAN so we are still using the development version and I am going to leave that workaround code there for now.

Tile size comparison. For each tile size test, I picked the tile that had the highest concentration of cells (outlined in red in the figure)

- 15x15, tile 79 = 10.239 minutes
- 20x20, tile 37 = 12.538 minutes
- 23x23, tile 24 = timeout
- 25x25, tile 22 = timeout
- 30x30, tile 19 = timeout

![image](https://user-images.githubusercontent.com/13220910/172250548-9b88accc-2d72-4d6a-8bc7-360d38a9f0dd.png)

PR #299, opened 2022-02-07
Redownload MN GCMs with grid and leap year fix

lindsayplatt:
More for #273. This uses the new grid from #297 and the new tile size from #298 to redownload the downscaled GCMs. First, this cut the download time down from 16 hours to 9 hrs 🎉. Second, the work described in https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1030234651 and https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1031981599 makes me confident that we have the grid correct. 

I also made adjustments to how we are using the `DateTime` and `time(day of year)` columns returned from GDP to create dates. The GCMs are built assuming that every year has 365 days, so if we just do `mutate(date = as.Date(DateTime)  + time(day of year))`, we end up mismatching our dates because R assumes you want to include Feb 29 if it is a leap year. It was quite the workaround to get something that correctly converts DateTime and the day of year (ranging from 0:364) into the appropriate day and skip any Feb 29ths. It was not aided by the fact that the `DateTime` is supposed to represent the first date of the current year but is quite messy when returned from GDP and sometimes you have values like `1984-12-31 23:15:03` representing the first day of 1985. For this reason, I had to implement some interesting code that figured out the actual first day for the data, then added one more day to dates after Feb 28 if it was a leap year in order to correctly shift dates to not include Feb 29 ever. I confirmed that this works by comparing one lake's data from my build of this pipeline to that of the [Winslow 2017 release](https://www.sciencebase.gov/catalog/item/57d30fd1e4b0571647d113e7) GCMs in https://github.com/USGS-R/lake-temperature-model-prep/issues/273#issuecomment-1032031775.

Unfortunately, the target for cleaning all the GCMs went from taking 2.5 minutes to just over 5 minutes after these leap year changes (not surprised due to the grouping etc that goes on). However, given that the download takes 9 hrs still, I felt like 5 minutes wasn't worth fretting over for now. Using `data.table` functions (or a hybrid approach with [`dtplyr`](https://dtplyr.tidyverse.org/)) could probably give us some gains, but I didn't want to spend too much time on optimizing now, especially with the looming deadline and without having it reviewed first.

### DIFF
--- a/7_drivers_munge/src/GCM_driver_utils.R
+++ b/7_drivers_munge/src/GCM_driver_utils.R
@@ -316,10 +316,7 @@ munge_notaro_to_glm <- function(in_file) {
   daily_data <- raw_data %>%
 
     # Create a column with the actual date
-    # `DateTime` has a POSIXct value for the first day of every year
-    # `time(day of year)` is a numeric value from 0-364 (365 for leap years)
-    mutate(date = as.Date(DateTime) + `time(day of year)`) %>%
-    select(-DateTime, -`time(day of year)`) %>%
+    convert_notaro_dates() %>%
 
     # Pivot to long format first to get cells as a column.
     pivot_longer(cols = -c(date, variable, statistic, units),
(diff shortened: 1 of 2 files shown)


7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -28,8 +28,8 @@ reconstruct_gcm_grid <- function(grid_params) {
 #' @param tile_dim single numeric value representing how many grid cells to
 #' group into a single tile. NOTE - sf_make_grid() can only make square grid
 #' polygons. The GCM grid is 217 cells wide by 141 high. With a `tile_dim` of
-#' 10 (tiles = 10 grid cells x 10 grid cells), tiles won't cover full height
-#' of the GCM grid and there will be 7 columns left out on the right and 1 left
+#' 15 (tiles = 15 grid cells x 15 grid cells), tiles won't cover full height
```
jread-usgs, line 31: glad this spend things up! Perhaps when we go multi-state we could get away with 20?

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -316,10 +316,7 @@ munge_notaro_to_glm <- function(in_file) {
   daily_data <- raw_data %>%
 
     # Create a column with the actual date
-    # `DateTime` has a POSIXct value for the first day of every year
-    # `time(day of year)` is a numeric value from 0-364 (365 for leap years)
-    mutate(date = as.Date(DateTime) + `time(day of year)`) %>%
-    select(-DateTime, -`time(day of year)`) %>%
+    convert_notaro_dates() %>%
```
jread-usgs, line 319: Agree with pulling this funky stuff out into a function!

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -401,3 +398,66 @@ validate_notaro_units_assumptions <- function(data_in) {
...
+  data_in %>%
+    # select(DateTime, `time(day of year)`, variable) %>%
+
+    ### Group data into the different big time chunks by finding and identifying ###
+
+    # This groups data at the points where 2000 is right before 2040, and 2040 is right before 2060
+    mutate(diff_days = abs(as.numeric(DateTime - lag(DateTime))/86400)) %>%
+    # Use a difference in days greater than one year to determine if a row is the
+    # start of a new time period
+    mutate(new_time_period = is.na(diff_days) | diff_days > 366) %>%
```
jread-usgs, line 425: wow. 😱

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -401,3 +398,66 @@ validate_notaro_units_assumptions <- function(data_in) {
...
+#' current year of data, e.g. the `DateTime` values for 1985 data are `1984-12-31 23:15:03`.
+#' @param data_in a data.frame with a least the columns `DateTime` (a POSIXct
+#' object with the first day of each year of data), `time(day of year)` (a numeric
+#' column with the day of year, ranging from 0:364), and `variable` (the downscaled
+#' GCM column containing data variables, e.g. `prcp_debias`).
+#' @value a data.frame with the columns `date`, `variable`, and any other column from
+#' `data_in` that is not `DateTime` or `time(day of year)`.
+convert_notaro_dates <- function(data_in) {
+  data_in %>%
+    # select(DateTime, `time(day of year)`, variable) %>%
```
jread-usgs, line 417: probably delete this `select` comment?

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -401,3 +398,66 @@ validate_notaro_units_assumptions <- function(data_in) {
...
+
+    ### For each time period + variable group, identify the earliest year in the period and use that to create a column of years ###
+
+    # Not able to just to `format(DateTime, "%Y")` to get the year because not all have a `DateTime` in the
+    # correct year, e.g. 1985 records have a DateTime of `1984-12-31 23:15:03`.
+
+    # First treat rows for the same variable and time period as a group
+    group_by(variable, new_time_period_i) %>%
+    # Determine the earliest year in the time period
+    mutate(earliest_year = as.numeric(format(min(DateTime), "%Y"))) %>% # mutate(year_1 = as.numeric(format(min(DateTime), "%Y"))) %>%
```
jread-usgs, line 437: Delete this `mutate` comment?

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -401,3 +398,66 @@ validate_notaro_units_assumptions <- function(data_in) {
...
+    # meaning the 59th day should be Mar 1 not Feb 29.
+    mutate(leap_adjustment = if_else(format(date_initial, "%m-%d") == "02-29", 1, 0)) %>%
+    # Use cumsum to make sure the leap_adjustment is used for every subsequent day of the year following a leap day
+    mutate(total_days_to_add = cumsum(leap_adjustment)) %>%
+    ungroup() %>% # Reset groups since they are no longer needed
+    # Using the initial date created, adjust the days to shift anything in a leap year that follows Feb 29
+    mutate(date_corrected = date_initial + total_days_to_add) %>%
+
+    # Keep only the variable and corrected date columns
+    select(date = date_corrected, matches("[0-9]+"), variable, statistic, units)
```
jread-usgs, line 462: Due to this PR requiring a lengthy download to run/verify, I'm not able to dig in and see what this `matches()` call is exposed to. But I think what this does is select date, variable, stat, units, and then any column that is _only_ numbers. Is that the case?

jread-usgs (approved):

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: Yes, I think so! I tested 20 cells per tile in #298 but it didn't make any difference for MN, so I just left as-is. I believe I left a not about increasing tile size when we scale up in #279

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: oops!

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: ack, oops again!

7_drivers_munge/src/GCM_driver_utils.R
lindsayplatt: Yes, correct! Any column that is _only_ numbers should be all the columns representing cells (since this is prior to pivoting). I suppose that comment is not comprehensive of what is kept by this `select()`

PR #302, opened 2022-02-14
Edit query map for use in data release - rename and add tile numbers

hcorson-dosch:
Modify code to generate `query_map_png`, so that the tiles are numbered on the map, and rename it to `tile_cell_map_png` to better describe what it shows. Given that we are [currently packaging](https://github.com/USGS-R/lake-temperature-process-models/blob/main/3_extract.R#L13-L30) up the prediction feather files by tile ([PR to make approach more flexible](https://github.com/USGS-R/lake-temperature-process-models/pull/26)), we can use this in the draft [data release](https://www.sciencebase.gov/catalog/item/6206d3c2d34ec05caca53071). As we finalize the data release, we may want to either revise this map or make a new version, since the tiles may not be relevant to how the final predictions are packaged.

### DIFF
--- a/7_drivers_munge/src/GCM_driver_utils.R
+++ b/7_drivers_munge/src/GCM_driver_utils.R
@@ -178,6 +178,10 @@ map_query <- function(out_file, lake_cell_tile_xwalk, query_tiles, query_cells, 
   grid_tiles <- grid_tiles %>%
     filter(tile_no %in% query_tiles)
 
+  tile_labels <- grid_tiles %>%
+    mutate(bbox=split(.,1:nrow(grid_tiles)) %>% purrr::map(sf::st_bbox)) %>%
+    unnest_wider(bbox)
+
   grid_cells <- grid_cells %>%
     filter(cell_no %in% query_cells) %>%
     left_join(lakes_per_cell)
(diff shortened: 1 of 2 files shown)


7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -178,14 +178,21 @@ map_query <- function(out_file, lake_cell_tile_xwalk, query_tiles, query_cells,
   grid_tiles <- grid_tiles %>%
     filter(tile_no %in% query_tiles)
 
+  tile_labels <- grid_tiles %>%
+    mutate(bbox=split(.,1:nrow(grid_tiles)) %>% purrr::map(sf::st_bbox)) %>%
```
lindsayplatt, line 182: I think you can use `tile_no` instead of `1:nrow(grid_tiles)` (that's the suggestion part of this comment). I've not seen `split()` + `purrr::map()` called _inside_ a mutate before. That is pretty nifty. I probably would have solved this by doing a `group_by()` + `mutate()` but the list part is necessary so I like having the `purrr::map()` call.

lindsayplatt (approved):

lindsayplatt:
Regardless of using this exact image in the data release or not, having the tile numbers added to the image is very useful.

PR #305, opened 2022-02-16
Only .gitignore 7_drivers_munge/out png files

hcorson-dosch:
Fix `.gitignore` file to only ignore `.png` files in `7_drivers_munge/out` and remove `.gitignore` line for `7_drivers_munge/in` that doesn't exist.


lindsayplatt (approved):
Wooo! World's smallest PR award 🏆

PR #288, opened 2022-01-21
Add univ mo data

padilla410:
This pull request completes the following:
* adds a cross walk to `2_crosswalk_munge` that matches the Univ MO state level ID system to NHDHR
* adds University of MO data to `all_dat_linked.feather` (closes #269)
* updates the Bull Shoals parser to crosswalk with the Univ. Mo crosswalk ([raw data](https://drive.google.com/file/d/11w6-LXCDSDCjipFYPxUgJyf7YB9BtXYR/view?usp=sharing) and [final xwalk](https://drive.google.com/file/d/1Shi5g5ig-J4PzM5ySHvodha52UIlw52h/view?usp=sharing)) (updates #218, will manually update after merge)
* updates `tidyr::gather` to `tidyr::pivot_longer` (closes #262)
* adds the Norfork sf rds to `getters.yaml` (closes #287)
* successfully runs the pipeline through `scmake("8_viz")`

The raw data from University of Missouri has 56,708 records. Approximately 33% of these records (18,599) make it into `all_coop_dat_linked.feather`. This relatively poor conversion rate is likely due to the point-in-polygon method used to crosswalk the state ID system used to the NHDHR system. This issue is documented in #289. Because of the issue, only data from three reservoirs makes it into the final data set: Mark Twain Res., Lake Stockton, and Truman Res.

Modifying the Bull Shoals [parser](https://github.com/USGS-R/lake-temperature-model-prep/blob/main/7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R) resulted in 5,780 additional records for 73 unique `DateTime` values.

Finally, while I was poking around in `crosswalk_coop_dat` I took a few minutes to convert `tidyr::gather` to `tidyr::pivot_longer`. `tidyr::gather` was superceded by `pivot_longer` a few years ago and I didn't want to wait until it was removed from the package.

After completing the above, we pick up 3 more PGDL lakes (compared to #284). Snapshot of `8_viz/out/lakes_summary_fig.html`:
image

PR #311, opened 2022-03-01
Adding MO US Army Corps of Engineers (USACE) data parser

### LINKED ISSUE #270: Bring in `Missouri_USACE_*_profiles` data
US Army Corps of Engineers (USACE) reservoir profile data from 2009-2021. Data includes the following:
* Site
* Date
* Time
* Depth (m)
* Temp (deg C)
* DO (mg/L)
* DO sat (%)

Other parameters (not included in all data sets):
* secchi
* Sp conductivity (uS/cm)
* pH
* Turbidity (NTU)
* PCY (cell/mL)

For a given year, the data is organized by reservoir, with multiple locations for each reservoir. Lat/longs are provided in a separate excel file called `Missouri_USACE_samplingLocations.xlsx`. The format of the data is optimized for "in-workbook" data analysis and some workbooks contain figures.

Note - there is likely sampling location overlap between this data set and the following data sets:
* Bull Shoals (PR #218)
* Univ of MO (#269)
* Navico (#219)

padilla410:
## Overview

This pull request completes the following:
 - adds a cross walk to `2_crosswalk_munge` that matches the MO USACE ID system (`mo_usace_id`) to NHDHR 
 - adds MO USACE data to `all_dat_linked.feather` (closes #270)

## Record Summary

After the initial parsing step, the raw data from MO USACE has 48,186 records. Approximately 55% of these records (26,643) make it into `all_coop_dat_linked.feather`. There are two main reasons for this:

1. Not all locations had matches in the spatial look up table. For example the lookup table (`Missouri_USACE_samplingLocations.xlsx`) has lat/longs for HD2, HD3, HD6, HD11 but not HD10 and HD15.
2. The point-in-polygon method doesn't always work well for reservoirs. For example, HD2 is in the look up table, but the coordinates are on land (first two images). In other cases, The reservoir is narrow and sinuous and the sampling location (WI9) is near the boundary of the waterbody (Wilson Lake) - this could result in a mismatch between the NHD polygon and the point of interest (third image).

image

image

image

## Results from `scmake("8_viz")`
After completing the above, we pick up 6 more PGDL lakes (compared to https://github.com/USGS-R/lake-temperature-model-prep/pull/288). Snapshot of `8_viz/out/lakes_summary_fig.html`:

image


7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+      purrr::map(readxl::read_xlsx, path = filepath, skip = subset_skip, col_names = T)
+
+    dat <- c(dat, dat_subset)
+  }
+
+  # remove sheets with no data
+  dat <- dat[sapply(dat, function(x) dim(x)[1]) > 0]
+
+  # some sheet names have extra whitespace and inconsistent case usage
+  names(dat) <- trimws(names(dat)) %>% toupper()
```
jread-usgs, line 245: Didn't know `trimws` was a function! handy

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+
+#' Validate column names
+#'
+#' MO USACE datasets have a wide variety of column names. This function
+#' checks for inconsistencies and fixes them. this function calls two
+#' helper functions: `run_name_checks` and `find_col_idx`
+#'
+#' @param ls_df a list of `data.frames` generated by reading in excel sheets
+#'
+validate_col_names <- function(ls_df) {
```
jread-usgs, line 325: this is a fairly generic function name and maybe only applies to a couple of spreadsheets? If so, you might want to add a word to the function name, since `validate_col_names` seems like a useful concept and maybe it would pop up elsewhere and cause a function name clash.

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+
+#' Identify column index
+#'
+#' Given a pattern, identify the column position of a column within a
+#' data.frame.
+#'
+#' @param df a data.frame
+#' @param pattern chr, pattern to match. Does not support partial matching.
+#'
+find_col_idx <- function(df, pattern) {
```
jread-usgs, line 393: I'd digging the modular code, but some of these really small functions may take things further than needed, esp the ones that are one line, like this. I think this one is used once and could appear here:

```r
site_col_idx <- find_col_idx(df, site_pattern)
names(df)[site_col_idx] <- 'site'
```
with 
```r
 site_col_idx <- which(apply(df, 2, function(x) any(grepl(site_pattern, x))))
 names(df)[site_col_idx] <- 'site'
 ```
 Instead, unless this function would be expected to be called elsewhere. 

Regardless, I found this one a little confusing, since I thought it was looking for names matches for column names, but it is actually looking at all of the values and returning TRUE (and then the index) for the columns that have these values, with partial matching (even though the doxy says it doesn't support partials)

```r
df[1:2, ]
# A tibble: 2 × 14
  ...1  date   time  `depth (m)` temp  conductivity       ph    turbidity          d.o.  d.o   `secchi depth`
  <chr> <chr>  <chr> <chr>       <chr> <chr>              <chr> <chr>              <chr> <chr> <chr>         
1 NA    MMDDYY HHMM  meters      øC    æS/cm              Units NTU                mg/l  % Sa… meters        
2 HT-1A 42109  1130  0.4         12.71 299.10000000000002 8.51  1.1000000000000001 10.6  99.1  NA           
```
☝️ I don't think "HT-1A" should have a partial match to "HT", but it does:
```r
apply(df[1:2, ], 2, function(x) any(grepl("HT", x)))
          ...1           date           time      depth (m)           temp   conductivity             ph 
          TRUE          FALSE          FALSE          FALSE          FALSE          FALSE          FALSE 
     turbidity           d.o.            d.o   secchi depth lake elevation        outflow          notes 
         FALSE          FALSE          FALSE          FALSE          FALSE          FALSE          FALSE 
```

My suggestion would be either to remove the function and put that code in the spot where it was calling it, or to change the name of this function to be more specific and also modify the documentation.

jread-usgs (approved):
Looks good. Some minor comments in here

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+  on.exit(unlink(unzip_dir, recursive = TRUE))
+
+  # Define variables needed to group and parse data -----
+  ## Define columns to keep
+  keep_cols_most_yrs <- c('Site', 'Date', 'Dep', 'Temp')
+  keep_cols_2009 <- c('Site', 'Date', 'Dep', 'Temp', '...9')
+  keep_cols_2018 <- c('Site', 'Date', 'Dep', 'Temp', '...4', '...5')
+
+  ## Define which years can be parsed together
+  most_yrs <- paste(c(2011:2017, 2019:2021), collapse = '|')
```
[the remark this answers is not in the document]
padilla410: Added a little more context to the comment

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+  if(!is.null(subset_sheets)) {
+
+    # sanity check
+    if(is.null(subset_skip)){stop("`subset_skip` must be specified if `subset_sheets` is specified")}
+
+    # remove `subset_sheets` from the data set
+    dat <- dat[!(names(dat) %in% subset_sheets)]
+
+    # read in the sheets that have to skip `subset_skip` rows before reading
+    dat_subset <- filepath %>%
```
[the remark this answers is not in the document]
padilla410: The two weirdos include pieces of descriptive information in the first row instead of the data we are looking for.  Here is a screenshot of `PT` from `Missouri_USACE_2010_profiles.xlsx`:
image

This causes `read_xlsx` to start reading in the first row instead of skipping it like it would for a "normal" sheet (screenshot of `HD` from the 2010 dataset):
image

I ran a modified version of the code you proposed above and compared the two data sets using `all.equal` to verify that the results of the two approaches match. Here is the updated snippet:
```r
dat_subset <- subset_sheets %>%
   purrr::set_names() %>% 
   purrr::map(readxl::read_xlsx, path = filepath, skip = subset_skip, col_names = T) # removed "set" and "subset_sheets"
```

I am going to update the function with your piece of code.

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
padilla410: Agreed, I will make this function name a bit more specific. Validating column names is a key piece of getting this data set to munge properly and is needed in almost every year of data provided.

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
padilla410: I think I am going to go with second option suggested:

> change the name of this function to be more specific and also modify the documentation

I don't have a good reason for why the doc on `find_col_idx` says it doesn't support partial matching. Especially considering that `run_name_checks` explicitly says that it does! 😬 

I know it's a tiny function, but I want to keep it because I am making _a lot_ of calls to the `apply` family of functions in this parser. I think the wrappers keeps it readable. Also, I definitely thought I was going to use this function more than once! 😄

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+    coerce_numeric_columns(.) %>%
+
+    # Site IDs are not not backfilled in most of the data sets
+    backfill_site_ids(.) %>%
+
+    # Removes all partially populated row created by the
+    # previous two munge steps
+    remove_incomplete_rows(.) %>%
+    convert_date(.) %>%
+    dplyr::bind_rows(.)
```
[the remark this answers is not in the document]
padilla410: Naming conventions for `sites` in this dataset are highly variable - both within years and between years _and_ when compared to the spatial data provided. Because the issues with names in the `site` column are so widespread I opted to clean up the `site` names on all datasets at the end of the parser ([link](https://github.com/USGS-R/lake-temperature-model-prep/blob/e82c8b9f1dd73a4715bbc0f17b43ea5ccc6e9861/7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R#L74-L77)). That being said, I am going to translate the that snippet of code into tidyverse as requested a bit earlier in the review ([link](https://github.com/USGS-R/lake-temperature-model-prep/pull/311#discussion_r817126683)).

I **think** I have answered your question, but please let me know if I have not! 😅

7a_temp_coop_munge/src/data_parsers/parse_mo_usace.R
```diff
@@ -0,0 +1,427 @@
...
+    dat_2018
+  )
+
+  # Light cleaning to match the sampling location xwalk table
+  data_clean$site <- gsub(" ", "", data_clean$site)
+  data_clean$site <- gsub("-", "", data_clean$site)
+  data_clean$site <- toupper(data_clean$site)
+
+  # naming convention clean-up for merging and xwalking
+  names(data_clean)[2] <- 'DateTime'
```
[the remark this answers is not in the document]
padilla410: done!

padilla410:
@jread-usgs I believe I have addressed all comments, but please take quick peek at the one conversation I have not marked as "resolved". I just want to make sure we have the same understanding on that one. 

If you give me a thumbs up, I'll merge this.

jread-usgs:
Good to merge @padilla410

PR #312, opened 2022-03-02
Adjust nc global attributes

### LINKED ISSUE #291: GCM Munging - finalize netCDF format before data release completion
In #252, we landed on providing the WGS84 coordinates for the grid cell centroids in the munged GCM netCDF files, in place of the projected coordinates (see #281). At the moment, we're also including the original GCM grid crs information and grid parameters in the global attributes of those netCDF files:
image

Before we complete the data release, we'll want to run this approach by Blodgett, as it may confuse users and make them think the coordinates provided in these netCDF files are in that CRS, rather than WGS84 (see discussion [here](https://github.com/USGS-R/lake-temperature-model-prep/issues/252#issuecomment-1022504277)).

hcorson-dosch:
This fixes #291 by adjusting the global attributes of the exported netCDF files and adding a new `source` attribute per Dave Blodgett's recommendations [here](https://github.com/USGS-R/lake-temperature-model-prep/issues/291#issuecomment-1050197255).

The resulting netCDF files look like this:
image


7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -59,8 +59,19 @@ generate_gcm_nc <- function(nc_file, gcm_raw_files, vars_info, grid_params, spat
 
   # Set up attributes for NetCDF that are independent of variables
   data_time_units <- "days since 1970-01-01 00:00:00"
-  data_attributes <- list('title' = global_att)
-  data_attributes <- append(data_attributes, grid_params) # add grid parameters to list of global attributes
+  source_info <- list('title' = sprintf("Notaro debiased downscaled GCM data - %s", gcm_name),
+                      'source' = paste("These data were derived from a source raster dataset of dynamically downscaled",
+                                       "global climate model (GCM) data generated by Michael Notaro (mnotaro@wisc.edu)",
+                                       "and others (see https://ccr.nelson.wisc.edu/dynamical-downscaling/index.php).",
+                                       "For internal USGS users, the source dataset can be accessed at this url:",
```
jread-usgs, line 66: I think we should leave out the internal data url, since it isn't likely to remain hosted there into the future.

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -59,8 +59,19 @@ generate_gcm_nc <- function(nc_file, gcm_raw_files, vars_info, grid_params, spat
 
   # Set up attributes for NetCDF that are independent of variables
   data_time_units <- "days since 1970-01-01 00:00:00"
-  data_attributes <- list('title' = global_att)
-  data_attributes <- append(data_attributes, grid_params) # add grid parameters to list of global attributes
+  source_info <- list('title' = sprintf("Notaro debiased downscaled GCM data - %s", gcm_name),
+                      'source' = paste("These data were derived from a source raster dataset of dynamically downscaled",
```
jread-usgs, line 63: "These data were derived from a source raster dataset of dynamically downscaled _and debiased_"

7_drivers_munge/src/GCM_driver_nc_utils.R
```diff
@@ -59,8 +59,19 @@ generate_gcm_nc <- function(nc_file, gcm_raw_files, vars_info, grid_params, spat
...
+                                       "global climate model (GCM) data generated by Michael Notaro (mnotaro@wisc.edu)",
```
jread-usgs, line 64: I think the primary use of this acronym is "General Circulation Model", but it is also used in this way. I think we should go with the primary definition of the acronym.

jread-usgs (commented):
Looks good. Minor feedback here in the text.

hcorson-dosch:
👍 Thanks for the review, Jordan!

Issue #310, opened 2022-02-28
Check on modeling status of MO reservoirs


lindsayplatt:
We need to check on the status of our ability to model some MO reservoirs and get that info to Craig. I think the  `8_viz/inout/lakes_summary.feather.ind` ​file has the info we need. I wrote the following code to find the site ids and then look at the summary table information. I thought maybe it was out-of-date, so I tried building it with `scmake('8_viz/inout/lakes_summary.feather.ind')` and `7_config_merge/out/nml_meteo_fl_values.rds.ind` rebuilt (seems to rebuild every time ... which should maybe be logged as a separate issue) and then `GLM_driver_files` started to build, so I stopped it! Handing off to @padilla410 to take this home and confirm.

```r
library(tidyverse)

priority_reservoirs <- c("Bull Shoals",
                         "Table Rock",
                         "Lake of the Ozarks",
                         "Mark Twain",
                         "Truman",
                         "Pomme de Terre",
                         "Stockton",
                         "Wappapello")

##### Find nhdhr ids for the priority MO reservoirs #####

lake_to_state_xwalk <- readRDS(scipiper::sc_retrieve('2_crosswalk_munge/out/lake_to_state_xwalk.rds.ind'))
gnis_name_to_nhdhr_xwalk <- readRDS(scipiper::sc_retrieve('2_crosswalk_munge/out/gnisname_nhdhr_xwalk.rds.ind'))

mo_lakes <- lake_to_state_xwalk %>%
  dplyr::filter(state == "MO") %>%
  dplyr::left_join(gnis_name_to_nhdhr_xwalk, by = "site_id")

# Note that some have multiple nhdhr ids
priority_reservoirs_ids <- mo_lakes %>%
  dplyr::filter(grepl(paste(priority_reservoirs, collapse="|"), GNIS_Name)) %>%
  dplyr::select(site_id, GNIS_Name)

##### Load summary data #####

lakes_summary <- arrow::read_feather(scipiper::sc_retrieve('8_viz/inout/lakes_summary.feather.ind'))

lakes_summary_mo <- lakes_summary %>%
  dplyr::right_join(priority_reservoirs_ids)

# Got these GLM vs PGDL criteria from the `lake_data_map.Rmd` file
lakes_glm <- lakes_summary_mo %>% dplyr::filter(zmax & kw & meteo) # I AM CONCERNED THIS IS NOT ACTUALLY CORRECT (not NAs in n_profiles)
[...]

PR #313, opened 2022-03-03
Update full 8_viz

jread-usgs:
I'm not sure why these would change. Perhaps the current version of data in the pipeline wasn't run all the way through. 

But since we've got a shared cache, I am checking these in so that others don't end up with surprising rebuilds due to clashes between local .ind files and gdrive files.


padilla410:
Hey Jordan, I am working on Lindsay's [request](https://github.com/USGS-R/lake-temperature-model-prep/issues/310#issuecomment-1058394852) from Issue #310 

As part of the work in #314 I will need to rebuild `8_viz` portion of the pipeline so that I can successfully complete the typo correction and verify that `zmax` is no longer `FALSE` using this snippet:

```r
arrow::read_feather('8_viz/inout/lakes_summary.feather') %>% filter(site_id == 'nhdhr_120032533')
# A tibble: 1 × 15
  site_id         GNIS_Name      n_obs n_profiles longitude latitude zmax  hypsography hypsography_dig… kw    kw_file meteo obs_category main_source Lake_Source
  <chr>           <chr>          <int>      <int>     <dbl>    <dbl> <lgl> <lgl>       <lgl>            <lgl> <lgl>   <lgl> <chr>        <chr>       <chr>      
1 nhdhr_120032533 Bull Shoals L…  1541         NA     -92.8     36.5 FALSE FALSE       FALSE            TRUE  FALSE   FALSE none         NA          ""     
```

Will I need to pull in the hash updates from this PR into my branch before I run `scmake('8_viz")`?

6_drivers_fetch.yml
```diff
@@ -22,7 +22,7 @@ targets:
       - "6_drivers/log/feather/6_fetch_new_cell_tasks.ind"
 
   replete_time_range:
-    command: c(t0=I(0), t1 = I(367700)) # for doing one timechunk: 2499
+    command: c(t0=I(0), t1 = I(367700)) # for doing one timechunk: 24999
```
padilla410, line 25: Is this simply a typo correction? I don't see any code changes - only a comment change

padilla410:
Also, the only reason I can think of for the hash update is that maybe we were leapfrogging each other at the end of the day yesterday.

padilla410 (approved):

6_drivers_fetch.yml
jread-usgs: correct. This is an artifact of realizing I still had my remote push set to upstream/canonical instead of my fork after the master->main rename. I made a tiny comment change and pushed the commit to see where it went. This is me undoing that tiny comment and it was added to this PR because it was too small to add elsewhere.

PR #315, opened 2022-03-03
Fix out skipnc feather

hcorson-dosch:
Make the `glm_ready_gcm_data_feather` back into a branched target so that the `out_skipnc_feather` target can map over it to build the intermediary feather files.

### DIFF
--- a/_targets.R
+++ b/_targets.R
@@ -178,8 +178,13 @@ targets_list <- list(
   # Combine the `file_out` elements of the gcm_ready_gcm_data_list branches
   # into a single vector of file targets
   tar_target(
+    glm_ready_gcm_data_paths,
+    unlist(glm_ready_gcm_data_list[grepl("file_out", names(glm_ready_gcm_data_list))])
+  ),
+  tar_target(
     glm_ready_gcm_data_feather,
-    unlist(glm_ready_gcm_data_list[grepl("file_out", names(glm_ready_gcm_data_list))]),
+    glm_ready_gcm_data_paths,
+    pattern = map(glm_ready_gcm_data_paths),
     format = "file"
   ),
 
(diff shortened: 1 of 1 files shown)


lindsayplatt (approved):
Yep, this is a targets gotchya that is not yet cemented in my brain :)

lindsayplatt:
BTW, here is the error that you get when you run the previous version. It is because We are trying to branch over a file target that wasn't created with a pattern. Hayley's fix here was the simplest way to fix it. You can also do it in a two-step pattern as shown in the `targets` documentation [here](https://books.ropensci.org/targets/dynamic.html#branching-over-files) or using a different function `tarchetypes::tar_files()`. Putting this here so that it shows up if we try to search the error in the future :)

```r
Error : Target out_skipnc_feather tried to branch over glm_ready_gcm_data_feather, which is illegal. Patterns must only branch over explicitly declared targets in the pipeline. Stems and patterns are fine, but you cannot branch over branches or global objects. Also, if you branch over a target with format = "file", then that target must also be a pattern.In addition: There were 14 warnings (use warnings() to see them)Error: callr subprocess failed: Target out_skipnc_feather tried to branch over glm_ready_gcm_data_feather, which is illegal. Patterns must only branch over explicitly declared targets in the pipeline. Stems and patterns are fine, but you cannot branch over branches or global objects. Also, if you branch over a target with format = "file", then that target must also be a pattern.Visit https://books.ropensci.org/targets/debugging.html for debugging advice.Run `rlang::last_error()` to see where the error occurred.
```

jread-usgs:
wow, this seems important, complicated, and confusing! I'm not sure I fully understand this gotcha, but perhaps it is something we'll want to work into the training?

lindsayplatt:
Good idea, @jread-usgs! See https://github.com/USGS-R/ds-pipelines-targets-3-course/issues/15

PR #324, opened 2022-04-04
rebuild GCM file ind

hcorson-dosch:
Rebuild the final `.ind` - noticed timestamp of `.nc` files was later in the day on March 3rd than then `7_drivers_munge/out/7_GCM_driver_files.ind` timestamp, so rebuilt the `.ind`


lindsayplatt (approved):
Any ideas as to why they were out of sync?

hcorson-dosch:
I did rebuild before merging #309, but I think I forgot to rebuild before merging #312

PR #319, opened 2022-03-17
Finalize hypso resampling

hcorson-dosch:
Fixes #293 

Add code to resample hypsography prior to creation of the `7_config_merge/out/nml_list.rds.ind` target.

This code is set up such that:
* Hypsography is only resampled for lakes with > 2 values in the H and A vectors (those lakes with additional bathymetric information beyond max depth)
* No duplicate elevation-area pairings are generated
* The resampled area values increase monotonically with elevation
* The elevation intervals to which the area values are resampled is determined based on lake depth, with an interval of 1m for deep lakes, 0.5m for medium-deep lakes, and 0.25m for shallow lakes. The current depth thresholds (8m, 5m, and 3m, respectively) were pulled from the values used to determine `max_layer_thick` [here](https://github.com/USGS-R/lake-temperature-model-prep/blob/main/7_config_merge/src/munge_lake_info.R#L149-L155) but can easily be adjusted
* The radaii rather than the area values are resampled with `approx()` and are then used to calculate the area values, per Lindsay's method used in [`lake-temperature-out`](https://github.com/USGS-R/lake-temperature-out/blob/main/2_process/src/calculate_toha.R#L302-L313)

PR #325, opened 2022-04-04
updates after a full build, pulling upstream

jread-usgs:
I built `8_viz` to see if I'd run into the mysterious rebuild issues captured elsewhere #321 and #322

But I didn't see any of those, but did end up with changes in the final viz steps and files. I'm not sure why, since Lindsay's PR went all the way through. Perhaps other data has been added since that hasn't gone through the whole pipeline until now?

image
But I'm current with Julie's map from #321 

Either way, I think this needs to be merged because it represents changes that would be on gdrive, so the inds and build files need to be kept current.


lindsayplatt (approved):
These hash changes are actually due to the hypso resampling from #319. We didn't build through `8_viz` in that one, so the changes to the hypso weren't completely propagated through the pipeline.

Issue #261, opened 2021-12-16
Include UNDERC dataset? 


jzwart:
I was looking at the data status map and noticed University of Notre Dame Environmental Research Center  (UNDERC) lakes aren't included in the dataset (green blocks in screen shot below) 
![image](https://user-images.githubusercontent.com/4825431/146416829-d61c5965-c250-4b93-97df-8ab1e0deae1e.png)

There are ~170 lakes in the [Solomon et al data set ](https://caryinstitute.figshare.com/articles/dataset/MFE_database_Data_from_ecosystem_ecology_research_by_Jones_Solomon_and_collaborators_on_the_ecology_and_biogeochemistry_of_lakes_and_lake_organisms_in_the_Upper_Midwest_USA/7438598) with max depth information, ~95 lakes with temperature observations. A bunch of other data including water clarity in there too

lindsayplatt:
@jread-usgs checking on how separate depth and temp are from the rest of the data to see how big of a lift it would be to include.

jread-usgs:
@jzwart I wonder if these are in LAGOS, which would be covered in #220 and seems like a priority. 

If not we could pull them in via that figshare dataset and contact Stuart prior to doing so.

jzwart:
My guess is that they aren't in LAGOS. @joneslabND @ctsolomon or @kaijagahm do you know if  [Solomon et al data set](https://caryinstitute.figshare.com/articles/dataset/MFE_database_Data_from_ecosystem_ecology_research_by_Jones_Solomon_and_collaborators_on_the_ecology_and_biogeochemistry_of_lakes_and_lake_organisms_in_the_Upper_Midwest_USA/7438598)  is in LAGOS?

jread-usgs:
I see some tasty lake data
<img width="1076" alt="image" src="https://user-images.githubusercontent.com/2349007/162060383-047f97e5-d9ac-4f35-8a86-005bd3782e63.png">

joneslabND:
We measure water color in the lab, but there is a lot of PAR profile data too. Happy to chat about the database sometime if that is helpful. Obviously. Zwart knows a lot about it already.

joneslabND:
[...]

Issue #176, opened 2020-07-23
figure out what is limiting us from including Isle Royale lakes


jread-usgs:
![image](https://user-images.githubusercontent.com/2349007/88332813-ca84aa80-ccf4-11ea-86fd-69ef3863b592.png)

We've got these lakes in the canonical set
```r
scipiper::scmake('1_crosswalk_fetch/out/canonical_lakes_sf.rds', remake_file = '1_crosswalk_fetch.yml') %>%
    readRDS %>% st_transform(crs = "+init=epsg:2811") %>% sf::st_simplify(dTolerance = 40) %>% 
    st_geometry() %>% plot(xlim = c(1154146.6, 1254146.6), ylim = c(434257.8, 464257.8))
```

Are we missing depth, drivers, clarity etc?

We should have temperature data on these lakes as there are loggers. But not sure where that gets linked

jread-usgs:
Gretchen found data by contacting NPS. I have uploaded the zip they shared into `3_params_fetch/in/ISRO_GLKN_Bathymetric_FinalData.zip.` and used `gd_confirm_posted('3_params_fetch/in/ISRO_GLKN_Bathymetric_FinalData.zip.ind')` to create the indicator file of the upload. 

I quick look at this: Looks like a geodatabase with 24 lakes. Some lakes are multipolygons, which I'd guess makes it easier to calculate contour areas, but some are multipolylines, which I'd guess means they have a break in the contour ring(s) somewhere, which makes it hard or we need to make assumpations. 

```r
st_layers('3_params_fetch/in//ISRO_GLKN_BathymetricData.gdb/')
Driver: OpenFileGDB 
Available layers:
                 layer_name     geometry_type features fields
1         Richie_1mContours Multi Line String       33      3
2    Chickenbone_1mContours Multi Line String       32      4
3        Sargent_1mContours Multi Line String       38      4
4         Harvey_1mContours Multi Line String       50      4
5         Beaver_1mContours Multi Line String       31      4
6          Beaver_CntrPolys     Multi Polygon       31      8
7      Ahmik_HalfM_Contours Multi Line String        6      4
8     George_HalfM_Contours Multi Line String       15      4
9  Feldtmann_HalfM_Contours Multi Line String       36      4
10      Siskiwit_2mContours Multi Line String      132      3
11         Siskiwit_6m_Poly     Multi Polygon       47      9
12         Desor_1mContours Multi Line String      101      4
[...]

Issue #167, opened 2020-04-12
mega issue: "fix" broken contours in IA, ND, and IN hypsography datasets


jread-usgs:
We have a number of contours that may need to be manually fixed/closed in QGIS or something else (not sure of a method to do it programmatically) where the contour lines end without connecting back to themselves. This causes us to either have to drop the lake or drop the contour ring. We may be losing some valuable lakes because of this.

jread-usgs:
Examples: broken polygon rings which don't close since these are linestrings. Also, casting to polygon may work, but accidentally subtract area because of node ordering (?) 
![image](https://user-images.githubusercontent.com/2349007/79117189-13b97b80-7d50-11ea-92bc-a1dcef646efa.png)

jread-usgs:
@hcorson-dosch no need to look at this for a long time, but just putting it on your radar as a topic to discuss in the project. 

Perhaps you have an approach you could recommend for this particular problem.

PR #327, opened 2022-04-12
UNDERC depth data and crosswalk, Isle Royale hypso data and crosswalk

jread-usgs:
I put both issue #261 and #176 in here. #261 is only partially resolved because the temperature data are not yet in the pipeline and the light attenuation data (additional processing of "PAR" profiles in the same dataset as the temperature) are required to gain more lakes. 

Despite the UNDERC dataset having 100+ lakes, we only gained one from that depth dataset. The others were either already in the pipeline or not in but still missing Kw/secchi data so not considered a "GLM" lake at this point, which is our lowest threshold for modeling. 

The Isle Royale (ISRO) dataset added more, since the pipeline seemingly already had temperature and secchi data for these sites (likely from WQP) and so adding hypsography resulted in ten new GLM lakes and eight new PGDL lakes. Also, since these lakes seem to be highly observed, it is worth noting that I only added the low hanging fruit contour lakes which could be easily turned into hypso. I didn't try to handle multilines so I'm not sure if issues like #167 will exist or if these could be cast into polygons without issue. We probably want to spin off another lower priority issue for the rest of the ISRO lakes. 

image


jread-usgs:
Oh yeah, a zoom in on the island 
image

for #176

1_crosswalk_fetch.yml
```diff
@@ -127,6 +128,16 @@ targets:
     command: slice_iadnr_contour(target_name,
       contour_ind = '1_crosswalk_fetch/out/iadnr_contours_sf.rds.ind')
 
+  1_crosswalk_fetch/out/isro_contours_sf.rds.ind:
+    command: fetch_isro_lakes(target_name,
+      use_geoms = I('Multi Polygon'),
+      gdb_filename = I('ISRO_GLKN_BathymetricData.gdb'),
+      zip_ind = '3_params_fetch/in/ISRO_GLKN_Bathymetric_FinalData.zip.ind')
+
+  1_crosswalk_fetch/out/isro_lakes_sf.rds.ind:
```
lindsayplatt, line 137: Should this get added under the overall makefile depends list for `1_crosswalk_fetch`?

1_crosswalk_fetch/src/fetch_crosswalk.R
```diff
@@ -32,6 +32,54 @@ process_wbic_lakes <- function(file_out = '1_crosswalk_fetch/out/wbic_lakes_sf.r
   saveRDS(wbic_lakes, file = file_out)
 }
 
+fetch_isro_lakes <- function(out_ind, zip_ind, remove_IDs = c(), gdb_filename, use_geoms){
```
lindsayplatt, line 35: Note that `remove_IDs` is not used in the function code at all

1_crosswalk_fetch/src/fetch_crosswalk.R
```diff
@@ -32,6 +32,54 @@ process_wbic_lakes <- function(file_out = '1_crosswalk_fetch/out/wbic_lakes_sf.r
...
+    dplyr::select(layer_name, depth_col)
+
+  data_clean <- purrr::pmap(use_layers, function(layer_name, depth_col){
+    # get "Beaver" from "Beaver_CntrPolys" etc:
+    lake_name <- strsplit(layer_name, '_')[[1]][1]
+
+    st_read(file.path(gdb_path, gdb_filename), layer_name, quiet = TRUE) %>%
+      sf::st_zm() %>% st_transform(crs = 4326) %>%
+      rename(geometry = Shape) %>%
+      mutate(site_id = sprintf('isro_%s', lake_name)) %>%
```
lindsayplatt, line 70: This is lowercase but later in `2_crosswalk_munge` you pass in `ISRO_` to `poly1_ID_name`. Does that matter?

7a_temp_coop_munge/log/7a_temp_coop_munge_tasks.ind
```diff
@@ -1,153 +1,154 @@
-7a_temp_coop_munge/tmp/1945_2020_All_MNDNR_MPCA_Temp_DO_Profiles.rds.ind: ab2f36cec6a8803f1aa28f0ebe4d1d47
```
lindsayplatt, line 1: I wonder why all of these hashes changed?

lindsayplatt (approved):
Ran `scmake('8_viz/out/lakes_summary_fig.html.ind')` and it did what I expected up until it started to build `7_drivers_munge/out/7_drivers_munge_tasks.ind`, so I stopped it.

1_crosswalk_fetch/src/fetch_crosswalk.R
jread-usgs: I don't think we're consistent on this in the repo...but the reason for this is that `isro_{asdfasdf}` is the ISRO_ID field value, while the `ISRO_ID` that appears as an input value for `poly1_ID_name` in the `crosswalk_poly_intersect_poly()` function is the name of the field/column. I think for the most part the field/column name for these other IDs are all in caps, and for the most part (but I'm sure there are exceptions) the _value_ in that column is prefixed with the lowercase of it.

lindsayplatt: Hmmm ok, I think [the value for the UNDERC is capitalized](https://github.com/USGS-R/lake-temperature-model-prep/pull/327/files#diff-4ae5528d8c4172ac5cc98d0b5259d450cf4be0ba079243f68e54dc33a295c859R311). Sounds like this isn't a hidden bug or anything, just more of a style issue.

PR #331, opened 2022-04-18
Resolve missing xwalks

padilla410:
This PR resolves a self-inflicted merge conflict and adds an update to `.gitignore` so that my `.Rproj` file will be ignored.


jread-usgs:
@padilla410 We may create some unintended issues by gitignoring the .Rproj files for each contributor to the repo. We've had these checked into the repo before, is there a need to remove it now?

jread-usgs:
This is now an empty PR (no net changes to files), so you can merge it if you'd like

Issue #328, opened 2022-04-14
Update NLDAS driver data for lakes footprint to 2021


lindsayplatt:
We would like the most recent NLDAS driver data for the MO reservoir GLM v3.0 modeling (trying to be complete by the end of June 2022) and later for the full footprint GLM v3.0 modeling (completed later). This pipeline is most efficiently run by @jread-usgs, so assigning to him! Please complete by May 5th if you can so that it is ready to be use by the initial MO GLM3 models.

lindsayplatt:
Note that this update should include a change to cell coverage so that we end up with more model-able lakes, per this comment https://github.com/USGS-R/lake-temperature-model-prep/pull/329#issuecomment-1102874572

PR #332, opened 2022-04-19
NLDAS reset to the 2020 runs, to support ongoing other tasks

jread-usgs:
To support @lindsayplatt 's current issues getting past this stage in the build.

I had partial work on #328 which was causing issues and now I feel confident I know how to complete that task w/o impacting others.


jread-usgs:
Merging my own PR here w/ Lindsay's :+1: via chat

PR #336, opened 2022-04-26
Update time

### LINKED ISSUE #328: Update NLDAS driver data for lakes footprint to 2021

jread-usgs:
image
^ big changes after updating NLDAS to the whole CONUS. 

Getting this in now because it has build/ind updates and will provide more context and explanation tomorrow. 

Fixes #328 but this included some clean-up in how we were dealing with NLDAS data as specified originally in #1


jread-usgs:
This PR moves away from the rectangle-based NLDAS driver file builds originally created in #1. What the prior solution was doing was setting a rectangle to pull NLDAS NetCDF files for all time (but chunked into files of 25k steps each), writing all of the time values to single feather files for each variable and cell (still at native hourly NLDAS resolution) and then combining the variables into a single daily .csv for each cell. This process worked fine but each time we wanted to update time or space we had to set the appropriate rectangle and build the files. 

The new method I've coded works mostly the same but instead of using multiple rectangles in a mosaic covering our study area, it uses the single CONUS footprint [here](https://github.com/USGS-R/lake-surface-temperature-prep/blob/main/4_drivers_fetch.yml#L85). The indicator file from that build now lands in our `in` [here](https://github.com/USGS-R/lake-temperature-model-prep/pull/336/files#diff-35f65dd900779ea8ffd40a9d3f7facc23135304b2864afcdd896bfc551f080f4) and the other things work mostly the same. 

But, since we're no longer moving along these rectangles, which were manageable to build the files, I'm now grouping cells spatially in task plans so that chunks of cells and chunks of driver files are written per task, as opposed to a GLM .csv _per task_ which was the old workflow and not scalable. So that's what the other main changes are. Then some clean-up I took care of when I was in there.

jread-usgs:
I'm not sure this makes sense for a normal PR review, but I'd be happy to do a "explain it while we walk through it" type review if that would work better. @lindsayplatt I will defer to you as to how you'd prefer review works on this one.

lindsayplatt:
Love seeing `n=0` next to "GLM, but no meteo"!! 🚀

Issue #320, opened 2022-03-29
`data_queue` folder clean up and review


padilla410:
I went through the [`data_queue`](https://drive.google.com/drive/u/0/folders/1jGdWewg7uPaMvUwLMWtvfTEwSyYzHQDi) folder cleaned out files/folders that have been processed by remained in the queue. Below is a summary of the remaining files. Most have temperature and some depth information.

For most of this data it looks like other subsets of this data have been processed. For example, the following data sets are all in the `in` folder:

* Bull_Shoals_Lake_DO_and_Temp.zip
*  Indiana_Glacial_Lakes_WQ_IN_DNR.xlsx
* LeechMain2018.csv
* Michigan_micorps_data.xlsx

However, as you can see in the table there is at least one more file available from the same group (e.g., MiCorps) or at the same location (e.g., Bull Shoals).

|File|Type|Files|Loc|Parse|Notes|
| ---      | ---       | ---      | ---       | ---      | ---       |
|Data_from_LAGOS|folder|54|IL, IN, ND, SD||Has temperature, no parser? Not found in 'in'. There is a lot in here (54 files) and some have temperature (see explainer file).|
|Lake_clarity_PCA|folder|24|MN|N|No temp found in data. Looks like secchi only (based on spot checks)|
|Long29016100|folder|7|MN||Has temp, reasonably organized. Long Lake, MN|
|temp_DO_PCA|folder|24|MN||DO and temp, many locs. Looks like it came from a nice DB|
|Ten_mile_lake_BRUCE_DOW_11041300|folder|3|MN||Has temp. 2-3 files spread across some subfolders. Very messy. Other Tenmile Lake files have been processed.|
|NorfolkReservoir_*_AR_ monthlyTempDO_2016-2020.xlsx|file|2|AR||Has temp, no parser. Wasn't part of initial norfork request.|
|IN_lakedata_IDEM.xls|file|1|IN|N|WQ only. No lake temperature.|
|IN_lakedata_IUSPEA_alldata.xls|file|1|IN||Temp data with depth and loc|
|Indiana_GlacialLakes_ TempDOprofiles_5.6.13.xlsx|file|1|IN||Multiple lakes, depth, temp|
|Indiana_GlacialLakes_ TempDOprofiles_5.6.13.xlsx|file|1|IN||Multiple IN lakes with temp. Waterbodies are named, but haven't run down locs|
[...]

Issue #289, opened 2022-01-24
MO data loss due to point-in-polygon crosswalk method


padilla410:
University of Missouri data (documented in issue #269) comes with a table that includes spatial info for the eight lakes in the data set:
Lake Name | Lake ID Number | County | Lat | Long
-- | -- | -- | -- | --
Bull Shoals Lake | 100 | Ozark | 36.5308 | -92.6047
Lake Ozark - Bagnell Dam | 149 | Miller | 38.2039 | -92.6236
Lake Stockton | 93 | Cedar | 37.6917 | -93.7667
Mark Twain Lake | 145 | Ralls | 39.5223 | -91.6467
Pomme De Terre Lake | 92 | Hickory | 37.9017 | -93.32
Table Rock Lake | 98 | Stone | 36.595 | -93.3108
Truman Res. | 89 | Benton | 38.2639 | -93.4083
Wapapello Lake | 30 | Wayne | 36.9311 | -90.2806

However, after [completing the steps](https://github.com/USGS-R/lake-temperature-model-prep#crosswalk-alternative-id-comes-from-point-based-spatial-data-such-as-monitoring-locations) to link `Lake ID Number` to `NHDHR` only four waterbodies have `NHDHR` matches:

site_id | Missouri_ID | Lake Name | County
 -- | -- | -- | --
nhdhr_120032533 | Missouri_100 | Bull Shoals Lake | Ozark
nhdhr_106716325 | Missouri_93 | Lake Stockton | Cedar
nhdhr_120032180 | Missouri_145 | Mark Twain Lake | Ralls
nhdhr_120031146 | Missouri_89 | Truman Res. | Benton

The missing reservoirs are not small:
* Wapapello - ~3,400 ha
* Pomme De Terre - 3165 ha
* Table Rock - 17,440 ha
* Lake Ozark - ~21900 ha

and therefore are likely in NHDHR, but the spatial info might not align well with NHDHR data.

The count of profiles for each lakes ranges from 12-16 unique sampling events.

jread-usgs:
Quick map of the location for the Bagnell Dam:
![image](https://user-images.githubusercontent.com/2349007/150846994-c0038af6-0281-4230-b6e5-cf91439e8684.png)

So this likely fell outside of the point-in-polygon analysis, which lost the mapping to this reservoir. Probably similar for the others too, I'd guess.

PR #337, opened 2022-04-27
Add Three Bull Shoals Datasets from Issue 320

padilla410:
This PR is a partial solution to #320 ([this](https://github.com/USGS-R/lake-temperature-model-prep/issues/320#issuecomment-1110015652) table) and adds data from the following cooperator files:
* 20190409 DATA with all depths.csv 
* Bull Shoals and LOZ profile data LMVP
* Temp_DO BSL * 

The data in these files comes from two locations: Bull Shoals Reservoir and Lake of the Ozarks. The Bull Shoals data was successfully added to the pipeline, while the Lake of the Ozarks data was not, due to the unresolved issue #289.

### A `case_when` question
I have a specific question for @jread-usgs. While working on these parsers I was trying to find spots to use regular expressions and `case_when`. Specifically, I tried six ways from Sunday to integrate [this](https://github.com/padilla410/lake-temperature-model-prep/blob/906e031ea730ce3136db12e37b84e91b228db3e0/7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R#L97-L100) snippet into [this](https://github.com/padilla410/lake-temperature-model-prep/blob/906e031ea730ce3136db12e37b84e91b228db3e0/7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R#L89-L94) portion of a `dplyr` pipeline.

Specifically, I thought this should work:
```r
# read in data and clean
clean <- list_sheets %>%
  purrr::map_df(~ readxl::read_xlsx(path = file_path, sheet = .x)) %>%
  purrr::discard(~all(is.na(.x))) %>%  # remove columns with all NA
  dplyr::filter(!is.na(Timestamp)) %>% # remove rows without timestamp
  dplyr::rename_with(~ new_col_names) %>%
  mutate(
    
    DateTime = case_when(
      class(DateTime) == 'character' ~ mdy_hm(DateTime) %>% as.Date,
      all(class(DateTime)) != 'character' ~ as.Date(DateTime),
    ),
    
    Timezone = c('CDT/CST'),
    temp = fahrenheit_to_celsius(temp),
    depth = -1 * depth,
    Missouri_ID = 'Missouri_100' # value from Univ MO xwalk (https://drive.google.com/file/d/11w6-LXCDSDCjipFYPxUgJyf7YB9BtXYR/view?usp=sharing)
  ) %>%
  dplyr::select(DateTime, Timezone, depth, temp, site, Missouri_ID)
```
The reason I need to have conditional formatting for `DateTime` is because the `DateTime` field reads as either `character` or `POSIXct` for each of the 14 files in `6_temp_coop_fetch/in/Temp_DO_BSL_MM_DD_YYYY.zip`.

### Results from scmake("8_viz")
Bull Shoals **BEFORE**
image

Bull Shoals **AFTER**
image

Summary numbers (no change compared to #333):
image


lindsayplatt:
@padilla410 I am reviewing this now. Looking at your question about the use of `case_when()`, based on your prompt for it, it sounds like what you have does not work and you expected it to. Do you have any other info about what isn't working? That would help me troubleshoot a working solution.

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -1,6 +1,113 @@
...
+parse_20190409_DATA_with_all_depths <- function(inind, outind) {
 
+  infile <- sc_retrieve(inind, remake_file = '6_temp_coop_fetch_tasks.yml')
+  outfile <- as_data_file(outind)
+
+  dat_raw <- readr::read_csv(infile)
+
+  # split date and time
+  dat_datetime_split <- stringr::str_split_fixed(dat_raw$Timestamp, " ", 2) %>%
+    as_tibble %>%
```
lindsayplatt, line 11: Do you not need `as_tibble()` here? Interesting!

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
```diff
@@ -1,6 +1,113 @@
...
+read_bullshoals_data <- function(file_path, new_col_names, keep_cols) {
+
+  # identify sheets with data
+  list_sheets <- readxl::excel_sheets(file_path) %>%
+    .[grepl('^Pt', .)]
+
+  # read in data and clean
+  clean <- list_sheets %>%
+    purrr::map_df(~ readxl::read_xlsx(path = file_path, sheet = .x)) %>%
+    purrr::discard(~all(is.na(.x))) %>%  # remove columns with all NA
```
lindsayplatt, line 86: WOW!! I didn't know about `purrr::discard()` or `purrr::keep()`. Can't wait to use those in the future!

lindsayplatt (approved):
Looks good. I know you were trying to solve your DateTime fix using tidyverse and it seems like `case_when()` did not have the outcome you expected. Not sure what was going on but here is a working example that I think matches the use case you have?

```r
x <- tibble(datetime = c("2022-01-01 00:00:00"))
y <- tibble(datetime = as.POSIXct(c("2022-01-01 00:00:00")))

do_conversion <- function(df) {
  df %>% mutate(datetime_converted = case_when(
    is.character(datetime) ~ "CONVERTED",
    !is.character(datetime) ~ "LEFT ALONE"
  ))
}


> x %>% do_conversion()
# A tibble: 1 x 2
  datetime            datetime_converted
  <chr>               <chr>             
1 2022-01-01 00:00:00 CONVERTED         
> y %>% do_conversion()
# A tibble: 1 x 2
  datetime            datetime_converted
  <dttm>              <chr>             
1 2022-01-01 00:00:00 LEFT ALONE     
```

7a_temp_coop_munge/src/data_parsers/parse_missouri_files.R
padilla410: Glad to hear! I'm working on replacing my dependence on the `apply` family with `purrr` so I'm definitely digging into the doc.

PR #344, opened 2022-05-31
Update of WQP data pull

jread-usgs:
We've lost some Secchi data (from 14,664 to 14,617 "GLM lakes") but gained temperature data (from 2,884 to 2,898 "PGDL lakes"). I'm not concerned about the data loss because this often means bad data were removed from STORET or an org stopped sharing. 

I did this data pull after convo with Lindsay where we figured now would be a good time since a ton of model configs were already going to be updated for Kw value given the landsat pull in #343


jread-usgs:
old on left, new PR on right: 
image

PR #346, opened 2022-06-07
Scaling GCM munge steps to full footprint

lindsayplatt:
Fixes #279 (but not #280). Ended up with tiles of 20x20 cells. With these updates, the full build took just over 15 hours (the download piece alone was 13, which was less than what it was for just MN locally). 

Looking for @hcorson-dosch to review:

- `_targets.R`
- `7_drivers_munge.yml`
- `7_drivers_munge/out/7_GCM_driver_files.ind`
- `build/status/N19kcml2ZXJzX211bmdlL291dC83X0dDTV9kcml2ZXJfZmlsZXMuaW5k.yml`

Looking for @jesse-ross to review all the others (which he mostly has already):
- `README.md`
- `.gitignore`
- `docker/*`
- `gcm_drivers.slrum`
- `gcm_download.R` (though, I would really like this to just become a line in the slurm script, but I don't know how)
- `launch-rstudio-container.slurm`

PR #347, opened 2022-06-09
Add figure to show missing cells in GCM driver data pull

lindsayplatt:
OK, so we scaled and have the full footprint GCMs (see #346 where we adjusted tile size & filtered to NML only & moved to HPC), but unfortunately, the footprint of the Notaro GCMs stops in the middle of the Dakotas. With how we implemented the code to substitute a cell with data for a cell without works, but it means that cells across the same latitude on opposite sides of the state have the same driver data. We need to think about what we want to do about this issue.

`7_drivers_munge/out/query_tile_cell_map.png`:
image

`7_drivers_munge/out/query_tile_cell_map_missing.png`:
image

### DIFF
--- a/7_drivers_munge.yml
+++ b/7_drivers_munge.yml
@@ -38,7 +38,7 @@ targets:
   # kick off a build). Takes 5 seconds to build if the targets pipeline
   # is up-to-date.
   7_GCM_driver_files_date:
-    command: c(I('2022-06-07'))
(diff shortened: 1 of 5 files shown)


lindsayplatt:
@hcorson-dosch making you a reviewer for my new map that is built based on the original one & adding the US Map. Not looking for an extensive check, but want some additional eyes.

PR #350, opened 2022-08-04
Only retain lake cell mapping for lakes within GCM footprint

hcorson-dosch:
This PR adds in logic to filter out any lakes that fall outside of the GCM footprint before generating the final lake to cell mapping in `adjust_lake_cell_tile_xwalk()`.

Previously, Lindsay's exploration showed that a number of queried cells in North and South Dakota were not returning data and were instead being assigned data from the easternmost parts of those states (see #347):
query_tile_cell_map_missing_ORIG

With this PR, those lakes that fall outside [the footprint of the cells that did return data (the GCM footprint)](https://github.com/hcorson-dosch/lake-temperature-model-prep/blob/limit-to-gcm-footprint/7_drivers_munge/src/GCM_driver_utils.R#L209-L218) are [dropped](https://github.com/hcorson-dosch/lake-temperature-model-prep/blob/limit-to-gcm-footprint/7_drivers_munge/src/GCM_driver_utils.R#L220-L222) before the crosswalk is generated and data source cells are assigned for cells that did not return data that are _within_ the GCM footprint:
query_tile_cell_map_missing

As before, the total # of queried cells that didn't return data is 342, but now only the 76 that fall within the GCM footprint are being assigned data from a nearby cell. 266 fall outside of the GCM footprint. The exported `lake_cell_tile_xwalk.csv` reflects this, with 266 fewer unique `spatial_cell_no` yet the same # of unique `data_cell_no` as before.

I checked and this resulted in 613 lakes being dropped from the `lake_cell_tile_xwalk_df`

### DIFF
--- a/7_drivers_munge/src/GCM_driver_utils.R
+++ b/7_drivers_munge/src/GCM_driver_utils.R
@@ -157,8 +157,9 @@ get_lake_cell_tile_spatial_xwalk <- function(lake_centroids, grid_cells, cell_ti
 
 #' @title Adjust the lake-cell-tile xwalk by matching lakes to the queried cells
 #' that returned data.
-#' @description Using lake centroids, this spatially matches lakes to query cells
-#' that returned data. If the cell that the lake falls within is not missing data,
+#' @description Using lake centroids, this spatially matches lakes that fall within the
+#' bounding box of the returned GCM data to query cells that returned data. For lakes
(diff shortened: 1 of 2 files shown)


lindsayplatt (approved):
Awesome! Thanks for updating the figure, as well!

PR #352, opened 2022-08-09
Add exception for lotw

### LINKED ISSUE #351: Lake of the Woods gets dropped in new GCM filter
We needed to exclude cells outside of the GCM footprint (#350), but that means that we dropped Lake of the Woods again 😢 I think we should add an exception to the code to fill this missing cell in with data.

image

hcorson-dosch:
This PR manually adds Lake of the Woods back into the lake subset after all lakes that fall outside of the GCM footprint (including LotW) are filtered out. The cell that it falls within is now assigned a nearby cell to use for data. I modified the map to show all queried cells, which reveals why the cell matched to the cell that contains Lake of the Woods is two rows away, as it was the closest queried cell that returned data:
query_tile_cell_map_missing

Fixes #351

### DIFF
--- a/7_drivers_munge/src/GCM_driver_utils.R
+++ b/7_drivers_munge/src/GCM_driver_utils.R
@@ -385,7 +385,7 @@ map_missing_cells <- function(out_file, lake_cell_tile_xwalk, cell_info, grid_ce
           legend.position="top") + 
     ggtitle("What cells are missing driver data?", 
             subtitle = paste(
-              sprintf("%s queried cells (hollow, light grey border) returned data", length(grid_cells_w_data)),
+              sprintf("%s queried cells (hollow, light grey border) returned data and have lakes present ", length(grid_cells_w_data)),
               sprintf("%s of those cells (red border) are being used to fill in missing driver data for %s cells", length(cells_being_used), length(gcm_cells_w_o_data)),
               sprintf("%s queried cells (in grey) fell outside of the GCM footprint (blue box)",length(grid_cells_w_o_data_outside_gcm_bbox)),
               sep='\n')) +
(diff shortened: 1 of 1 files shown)


7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -217,10 +217,17 @@ adjust_lake_cell_tile_xwalk <- function(spatial_xwalk, lake_centroids, grid_cell
     st_as_sfc() %>%
     st_as_sf()
 
+  # pull out Lake of the Woods. It falls just outside of the bounding box of returned data,
+  # but has lots of data, so we want to keep it in our lake subset
+  lotw <- filter(lake_centroids, site_id=='nhdhr_123319728')
```
lindsayplatt, line 222: hehe this made me chuckle because it is so close to "LOTR" for Lord of the Rings 😄

lindsayplatt (approved):

7_drivers_munge/src/GCM_driver_utils.R
```diff
@@ -374,7 +385,8 @@ map_missing_cells <- function(out_file, lake_cell_tile_xwalk, cell_info, grid_ce
           legend.position="top") + 
     ggtitle("What cells are missing driver data?", 
             subtitle = paste(
-              sprintf("%s cells are being used to fill in missing driver data for %s cells", length(cells_being_used), length(gcm_cells_w_o_data)),
+              sprintf("%s queried cells (hollow, light grey border) returned data", length(grid_cells_w_data)),
```
lindsayplatt, line 388: Might reword to make it clear what we mean by "returned data" 
```suggestion
              sprintf("%s queried cells (hollow, light grey border) returned data and have lakes present ", length(grid_cells_w_data)),
```

7_drivers_munge/src/GCM_driver_utils.R
jordansread: 👋 DWALLIN

hcorson-dosch: Haha I kept thinking the same thing when I was coding it!

PR #354, opened 2022-08-18
Add the wind transform code

### LINKED ISSUE #353: Implement wind transform function for GCMs
After a deep dive into reasons behind some of the [warm bias we were seeing in GCM-driven GLM output](https://github.com/USGS-R/lake-temperature-process-models/discussions/46), we had a hunch that lower windspeeds in the GCMs than are present in NLDAS may be the issue. We did a test by adding 1.5 m/s to the windspeeds just to see how that would impact them and saw a significant change. These outcomes have encouraged us that correcting windspeeds would help with this bias.

Jordan has past experience with applying a transform function to do this statistical adjustment. The idea is that the distribution of GCM windspeeds in the contemporary time period should be similar to those in the NLDAS. Once that transform function is created, it can be applied to the future time periods.

lindsayplatt:
Fixes #353. I tested this munge code with the first 5 GCM raw feather files and the code does function! Here is a comparison of before (red) and after (blue) the transform for the same dates & GCMs:

image

@hcorson-dosch once this is merged, you can pull it into the `main` branch on Tallgrass and kick-off the re-builds.


jordansread (approved):
Confirmed this is the same method to apply the transform.

PR #356, opened 2022-08-23
Reorder data after wind transform

lindsayplatt:
The wind transform function added in #354 reordered the columns and that caused downstream issues in the `lake-temperature-process-models` repo because it assumes the dates are in order and GLM models failed. This should fix that!

I have already run `tar_make()` with these new changes on Tallgrass, so the nc files on Caldera should be updated now.


hcorson-dosch (approved):
Looks good! thanks for regenerating the nc files!

hcorson-dosch:
Note: Added a PR to your branch to add in the updated scipiper files: https://github.com/lindsayplatt/lake-temperature-model-prep/pull/4

Issue #357, opened 2022-08-23
Suspect bathy for two WI lakes


lindsayplatt:
Jacob noticed that there are two lakes in our lake metadata CSV on ScienceBase with > 200 m depths, while their depths according to the WI DNR website are 18 ft and 29 ft. I spent a few minutes investigating and it appears that the bathymetry data provided for these two sites via `3_params_fetch/in/WBIC_hypsos_lakeattributes.zip` have very deep depths - 228 and 383 (assumed meters), respectively. The only thing I can think is that they are either wrong OR recorded in inches (which would give 19 ft and 32 ft, respectively).

This is sort of related to #227, but most other depths in the dataset are reasonable (see histograms below)

```r
library(dplyr)
library(ggplot2)

# Quick summary of lake depths in the pipeline
munged_lakedepths_fn <- scipiper::gd_get('7_config_merge/out/nml_lake_depth_values.rds.ind')
munged_lakedepths <- readRDS(munged_lakedepths_fn)

munged_lakedepths %>% 
  mutate(is_pretty_deep = ifelse(lake_depth >= 100, "Depths >= 100", "Depths < 100")) %>% 
  ggplot(aes(lake_depth)) + geom_histogram() + facet_wrap(vars(is_pretty_deep), nrow=2, scales = "free")
```

![image](https://user-images.githubusercontent.com/13220910/186214508-8c9d96d4-52ec-4d30-ba4c-e5a6a9e09bd3.png)

Here are the  two lakes with depths > 200 m in the `7_config_merge/out/nml_lake_depth_values.rds` file
```r
# Jacob pointed out that two lakes have depths > 200 m in our SB metadata file (both are WI lakes)
very_deep_lakes <- munged_lakedepths %>% filter(lake_depth > 200)
very_deep_lakes

# A tibble: 2 x 2
# Rowwise: 
  site_id                                      lake_depth
  <chr>                                             <dbl>
1 nhdhr_{D8A9DD23-94AD-4C15-B242-FE519B5EDF8C}       228.
2 nhdhr_70333995                                     383.
```

Are those depths coming from the input WBIC depth file? No, the depths in `4_params_munge/out/wbic_depths.rds` match the depths listed on the WI DNR website
```r
# Take a look at those two lake depths from their input depths file
wbic_lakedepths_fn <- scipiper::gd_get('4_params_munge/out/wbic_depths.rds.ind')
[...]

PR #358, opened 2022-08-24
Build update with hypo fix

jordansread:
Followed plan in #357 and these are the build artifacts. 

Worked as expected (lost two hypsos, but no lost modelable lakes):

image


lindsayplatt (approved):
I guess these two lakes didn't have tons of data because the `Lots of data, no hypso in pipeline` counts didn't increase by two.

PR #363, opened 2023-03-17
Fix wqp file name bug in 7b_temp_munge step

padilla410:
## Overview

Our metadata reviewer for the [lstm data release](https://github.com/USGS-R/lake-temp-lstm-static-data-release/issues/37) found a bug in this pipeline when she ID'd many NA values in the output of `7b_temp_merge/out/temp_data_with_sources.feather`.

There were two issues going on : (1) I needed to get back into the metadata file and add info for some cooperators; and (2) there was no way to partial match file names for `wqp` tagged data. This PR fixes that bug in `add_source_ids` (within `7b_temp_merge/src/munge_source_ids`.

All of the additional files came along for the ride when I re-ran the pipeline.

Closes USGS-R/lake-temp-lstm-static-data-release#37

## Verification

Running this script should result in zero rows in rows in `na_temp_dat-with_src`

```
library(arrow)
library(tidyverse)

# Checkout the offending file - data with missing `source_id` values
temp_dat_with_src <- read_feather("7b_temp_merge/out/temp_data_with_sources.feather")
na_temp_dat_with_src <- temp_dat_with_src |> filter(is.na(source_id))
```


7b_temp_merge/src/munge_source_ids.R
```diff
@@ -39,15 +39,27 @@ gd_download_and_indicate <- function(source_metadata_loc, out_ind) {
...
+                       TRUE ~ file_name
+                       )
+           )
+
   dat_source <- left_join(dat, metadata, by = 'file_name') %>%
     dplyr::select(site_id, date, depth, temp, source_id)
 
+  # fix wqp files
+  dat_source <-
+
```
lindsayplatt, line 62: Looks like this may have accidentally been committed?
```suggestion
```

lindsayplatt (approved):

